PR #21, opened 2021-09-10
logjac for non-square matrices

cscherrer:
```julia
function logjac(x::AbstractMatrix) 
    (m,n) = size(x)
    m == n && return first(logabsdet(x))

    # Equivalent to sum(log, svdvals(x)), but much faster
    m > n && return first(logabsdet(x' * x)) / 2
    return first(logabsdet(x * x')) / 2
end
```

@sethaxen @mschauer does this look right?

In the SVD `A = U * Diagonal(S) * Vt`, the `U` and `Vt` are orthogonal. If these are isomorphisms or embeddings, they don't affect the geometry, so only the singular values matter. So I think this works as long as `m >= n` for sigma or `m <= n` for omega. But I want to be sure I'm not missing something.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "MeasureBase"
 uuid = "fa1605e6-acd5-459c-a1e6-7e635759db14"
 authors = ["Chad Scherrer <chad.scherrer@gmail.com> and contributors"]
-version = "0.4.2"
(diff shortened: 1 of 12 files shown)


cscherrer:
https://github.com/cscherrer/MeasureTheory.jl/pull/160#issuecomment-917317369

PR #30, opened 2021-12-03
Dev

cscherrer:


oschulz:
This is ready to try out, right?

cscherrer:
Yes. There might be some small adjustments to improve type inference, but from this point it should be pretty minor

PR #31, opened 2021-12-09
Refactoring

cscherrer:


cscherrer:
Getting pretty close, I think the biggest thing left is cleaning up the three-argument `logdensityof`

cscherrer:
I don't think I'll get through this right away (I have family in town soon). But I want to add some notes on the plan for this...

Say you have measures
- `mu::M` with `basemeasure_depth(mu) == m`
- `nu::N` with `basemeasure_depth(nu) == n`

We want to do density recursion on each of these. **But** if at any point there's an explicit method between them, we should prefer that. We really don't want this to be a quadratic thing, but I think we could constrain any explicit methods to be between measures of the same depth. With that, we could
1. Walk the greater-depth measure toward its root, taking the number of steps to make the resulting depths equal
2. Walk both together, from there, so we can check for explicit methods as we go.

PR #35, opened 2021-12-22
remove dead code

vtjnash:
Functions do not (always) have instances. And it seems unused.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -52,10 +52,4 @@ struct NonIterable end
 isiterable(::Type{T}) where {T} =
     static_hasmethod(iterate, Tuple{T}) ? Iterable() : NonIterable()
 
-functioninstance(::Type{F}) where {F<:Function} = F.instance
-
-# See https://github.com/cscherrer/KeywordCalls.jl/issues/22
-@inline instance_type(f::F) where {F<:Function} = F
-@inline instance_type(f::UnionAll) = Type{f}
-
 using MLStyle


cscherrer:
We do use this in the `dev` branch. I'll describe the problem here in case you see other ways to address it.

MeasureTheory is built around the concept of log-densities relative to some base measure. "Base measure" is not a canonical thing, but we need to make it canonical to work cleanly with `DensityInterface.logdensityof`. So starting with functions

```julia
logdensity_def(d::AbstractMeasure, x) :: Real
basemeasure(d::AbstractMeasure) :: AbstractMeasure
```
we need to get to a function like
```julia
function logdensityof(d, x)
    ℓ = 0.0
    b = basemeasure(d)
    while b != d
        ℓ += logdensity_def(d, x)
        (d, b) = (b, basemeasure(b))
    end
    return ℓ
end
```
But this has lots of overhead, because the loop is dynamic. It's much faster if we can know how many steps we'll need, and generate the loop structure statically. This leads to the idea of each measure having a static "depth".

Initially I was just making this something the user has to provide. But that wasn't composing so well, so I introduced a type-level version of `basemeasure`. This is called `tbasemeasure_type`. This is like `basemeasure_type` (i.e., the type of the base measure) with a leading `t` as suggested by jzr in Discourse (https://discourse.julialang.org/t/multiple-dispatch-value-and-type-methods/72901), to indicate that this takes a type argument.

This was working great until I got to Generators. Unlike arrays, there's no type parameter for the `eltype`, and `eltype` itself is much wider than it could be. I think this may also be a problem for things in Base.Iterators.

This seemed like a bug, so I filed
https://github.com/JuliaLang/julia/issues/43508#issuecomment-999079156
But that seems to not be the case.

I've explored making my own `_eltype` with the behavior I want, but now I'm wondering if it could make more sense to make a new `Base.Generator`-like type that has the type parameter I need.

PR #39, opened 2022-03-29
cleanup

cscherrer:
- Get `JET.report_package(MeasureBase)` to pass
- Drop the `partialstatic` stuff, fun idea but didn't seem to help much
- Add dependencies for `LogarithmicNumbers` and `StatsFuns`
- More docstrings
- Move `ConditionalMeasure` (`|`) here from MeasureTheory 
- Improvements for `superpose`
- Some minor tidying up


src/density.jl
```diff
@@ -32,7 +32,7 @@ function 𝒹(μ::AbstractMeasure, base::AbstractMeasure)
     return Density(μ, base)
 end
 
-logdensityof(d::Density, x) = logdensityof(d.μ, x) - logdensityof(d.base, x)
+logdensityof(d::Density, x) = logdensity_rel(d.μ, d.base, x)
```
mschauer, line 35: 👍

src/MeasureBase.jl
```diff
@@ -100,16 +127,19 @@ include("density.jl")
...
 using MeasureBase:basemeasure_depth, proxy
 using MeasureBase: insupport
-@reexport using Test
 
 export test_interface
 export basemeasure_depth
 export proxy
 export insupport
 
+@reexport using Test
```
theogf, line 141: Why does this need to be reexported?

src/MeasureBase.jl
```diff
@@ -100,16 +127,19 @@ include("density.jl")
 module Interface
```
theogf, line 127: This should maybe get its own file `Interface.jl` ?

src/combinators/conditional.jl
```diff
@@ -0,0 +1,8 @@
+struct ConditionalMeasure{M,C} <: AbstractMeasure
```
theogf, line 1: Could you add a docstring explaining what it does and how/when to use it?

src/density.jl
```diff
@@ -103,7 +103,26 @@ Define a new measure in terms of a log-density `f` over some measure `base`.
...
 # TODO: `density` and `logdensity` functions for `DensityMeasure`
 
-@inline function logdensityof(μ, x)
+"""
+    logdensityof(m::AbstractMeasure, x) 
+
+Compute the log-density of the measure `m` at `x`. Density is always relative,
+but `DensityInterface.jl` does not account for this. For compatibility with
+this, `logdensityof` for a measure is always implicitly relative to
+`rootmeasure(x)`.
```
theogf, line 112: If there is a docstring for `rootmeasure` maybe add a doc reference.

theogf (commented):
Here are my nitpick comments :) It's all I can do I as I don't fully understand all the details of the package yet.

src/MeasureBase.jl
cscherrer: Good idea, that will be clearer

src/MeasureBase.jl
cscherrer: Hmm, yeah maybe we don't need to export it. I'll make it just `using Test`

src/density.jl
cscherrer: I haven't used these before. Do you mean like `[rootmeasure](@ref)`?

theogf: Yes, the syntax is 
```
[`rootmeasure`](@ref)
```
but I think you can keep `rootmeasure(x)` and write it as
```
[`rootmeasure(x)`](@ref rootmeasure)
```

PR #43, opened 2022-05-02
Some convenience functions for WeightedMeasures

theogf:
This include having `length` returning the right thing, defining a fallback `logweight` for all measures and being able to sample from a weighted measure.


cscherrer:
Thanks @theogf . Can you help me understand downstream applications of these? My main concern is to be sure we can keep the semantics consistent.

theogf:
I think `length` and `rand` are pretty self explicit. For `logweight` it makes sense to me that a `Measure` which is not weighted should have a weight of `1`?
About the use, well it allows to not use API (like `length`) on weighted measures as well, but more specifically I would use for the sampling on superposed measure I mentioned [here](https://github.com/cscherrer/MeasureTheory.jl/issues/191#issuecomment-1114893270)

cscherrer:
I'm confused, because I don't think of measures as ever having a `length`, and I don't know what this would mean. Would it be the number of components of a product or superposition? Or the length of a `testvalue`? In any of these cases, why not just specify that directly?

Kind of the same issue with `logweight`. The name makes it sounds like it would be intrinsic to the measure itself, but this seems to instead look at the concrete representation. So for example, if `logweight` is defined for all measures, I'd think
```julia
logweight(3 * (Normal() + StudentT(5))) == log(6)
```
(the log of the mass of the measure)
But if I'm reading this correctly, it would instead give `log(3)`.

If it's going to not be composable, maybe we should call it `_logweight`, to show it's an internal and very concrete thing?

theogf:
> I'm confused, because I don't think of measures as ever having a length, and I don't know what this would mean

It probably should not be called `length` indeed! What I understood implicitly was the dimension of the variable domain.

>  The name makes it sounds like it would be intrinsic to the measure itself, but this seems to instead look at the concrete representation

I see what you mean, and your example is great. I will change accordingly

cscherrer:
> But that also means that the current `logweight` behavior is wrong right?

Haha, yeah I guess so. I don't see anywhere we use that, maybe we just drop it?

test/combinators/weighted.jl
```diff
@@ -0,0 +1,20 @@
...
+    μ = @inferred w * μ₀
+    @test μ == WeightedMeasure(log(w), μ₀) == weightedmeasure(log(w), μ₀)
+    @test μ isa WeightedMeasure
+    @test _logweight(μ) == log(w)
+    @test _logweight(w * μ) == 2 * log(w)
+    @test rand(MersenneTwister(123), μ) == rand(MersenneTwister(123), μ₀) 
+    x = rand()
+    @test logdensity_def(μ, x) == log(w)
+    @test logdensityof(μ, x) == logdensityof(μ₀, x)
+end
```
cscherrer, line 20:
```suggestion
end

```

theogf:
Btw, back to the `length`, `length` on a `ProductMeasure` returns the dimensions of the input, so having `length` on `WeightedMeasure` helps passing this information.

theogf:
Ah but it actually just return the number of marginals, not in the input dimensionality.
There should probably be another function for that. In the mean time I will just remove it.

theogf:
@cscherrer Anything else you would like to see changed?

cscherrer (approved):

cscherrer:
Looking good, thanks @theogf !

PR #44, opened 2022-05-02
Smarter CI

theogf:
Do not trigger CI twice when pushing on PR


cscherrer:
Great! Thanks @theogf

PR #46, opened 2022-05-03
Fix a few issues with functions on superposition

theogf:
This fixes a few issues where functions where specialized for only 2-tuples of measure.

Sorry there were a few autoformatting


src/combinators/superpose.jl
```diff
@@ -76,13 +76,21 @@ oneplus(x::ULogarithmic) = exp(ULogarithmic, log1pexp(x.log))
     return dμ_dα / oneplus(dβ_dα) + dν_dβ / oneplus(dα_dβ)
 end
 
+function density_def(s::SuperpositionMeasure, x)
+    error("Not implemented")
+end
```
cscherrer, line 81: What problem do you see with
```julia
density_def(μ, x) = exp(logdensity_def(μ, x))
```
?

src/combinators/superpose.jl
```diff
@@ -76,13 +76,21 @@ oneplus(x::ULogarithmic) = exp(ULogarithmic, log1pexp(x.log))
...
+@inline function logdensity_def(
+    μ::T,
+    ν::T,
+    x::Any,
+) where {T<:(SuperpositionMeasure{Tuple{A,B}} where {A,B})}
     if μ === ν
         return zero(return_type(logdensity_def, (μ, x)))
     else
-        return logdensity_def(μ,x) - logdensity_def(ν, x)
+        return logdensity_def(μ, x) - logdensity_def(ν, x)
```
cscherrer, line 93: Is this just a formatting change? That's fine, just want to be sure.

We should also drop the `::Any`, since it's not needed

src/combinators/superpose.jl
```diff
@@ -94,7 +102,11 @@ end
     return logaddexp(logdensity_rel(μ, β, x), logdensity_rel(ν, β, x))
 end
 
-@inline function logdensity_def(s::SuperpositionMeasure{Tuple{A,B}}, β::SuperpositionMeasure, x) where {A,B}
+@inline function logdensity_def(
+    s::SuperpositionMeasure{Tuple{A,B}},
+    β::SuperpositionMeasure,
+    x,
+) where {A,B}
```
cscherrer, line 109: Also just formatting?

src/combinators/superpose.jl
```diff
@@ -107,7 +119,9 @@ end
 
 @inline logdensity_def(s::SuperpositionMeasure, x) = log(density_def(s, x))
 
-basemeasure(μ::SuperpositionMeasure) = superpose(map(basemeasure, μ.components)...)
+basemeasure(μ::SuperpositionMeasure{Tuple{A,B}}) where {A,B}  =
+    superpose(map(basemeasure, μ.components)...)
+basemeasure(μ::SuperpositionMeasure) = superpose(map(basemeasure, μ.components))
```
cscherrer, line 124: This makes me wonder if we should go even farther in this direction. Should we be doing `...` at all for superpositions? If so, I think we're missing a method like
```julia
superpose(m1, m2, ms...) = ...
```
But maybe we should just be doing this within a Tuple?

cscherrer (commented):
Thanks @theogf , I've added a few comments/questions

src/combinators/superpose.jl
theogf: Yeah sorry, I have this autoformatter on and it just changed things unwillingly :sweat_smile: 
If you are open to it, we could have one big PR to follow a formatting style and there is an action to automatically make formatting suggestions on PR.

src/combinators/superpose.jl
theogf: Yes!

src/combinators/superpose.jl
theogf: What do you think of 
```julia
function superpose(μ, μs...)
    if all(==(μ), μs)
        return weightedmeasure(log(length(μs) + 1), μ)
    else
        return superpose((μ, μs...))
    end
end
```

src/combinators/superpose.jl
cscherrer: But it can take a `Tuple{A,B}`. I think we're talking past each other - do you have an example that currently fails?

src/combinators/superpose.jl
cscherrer: By itself I like it, but we need to think through the context and the dispatch patterns that lead us to this.

Users can build a `SuperpositionMeasure` in a few ways. We need `a + b` and `sum((a,b))` to result in the same thing. I think we should have one method to handle both cases. Whether that's a method on Tuples or a `...` method mostly comes down to performance and which is more natural to code.

There's a fair amount more to do for general superpositions. In the end we need a data structure that encodes how the base measures are related, storing everything in a tree structure. That will make it easy to compute things efficiently, since it will let us share log-density computations. It will also make it easier to fold more measures into an existing superposition.

We'll also need to build the `logmass` API we've talked about before. I'll turn this comment into a new issue, so we can discuss there.

theogf: > We need a + b and sum((a,b)) to result in the same thing. I think we should have one method to handle both cases.

I think the only thing to handle is the `a::SuperpositionMeasure + b::Measure` dispatch (two versions one for tuple and one for vectors).

cscherrer: Ahh ok, so as a start, something like
```julia
superpose(a::AbstractMeasure, b::AbstractMeasure, ms...) = superpose(superpose(a, b), ms...)
superpose(a::SuperpositionMeasure, b::AbstractMeasure) = ...
superpose(a::AbstractMeasure) = a
```

src/combinators/superpose.jl
theogf: I think this is configurable in a Style file

src/combinators/superpose.jl
theogf: Yeah sure running `logdensityof(SuperpositionMeasure([Dirac(0), Dirac(2.0)]), 2.0)`

cscherrer:
I want to summarize something here that was confusing me a little...

We typically work in terms of `logdensity_def`, which is why we have methods
```julia
density_def(μ, ν::AbstractMeasure, x) = exp(logdensity_def(μ, ν, x))
density_def(μ, x) = exp(logdensity_def(μ, x))
```
But superpositions are more naturally expressed in "linear space", which is why I had added
```julia
@inline logdensity_def(s::SuperpositionMeasure, x) = log(density_def(s, x))
```

But as @theogf points out, this sets up a mutual recursion, so now the default case will blow the stack. So in this case, we're really depending on `density_def` being defined. To make this safer and more user-friendly, we should have a default method that throws an informative error, since this is more useful than a stack overflow. This leads us to
```julia
function density_def(s::SuperpositionMeasure, x)
    error("Not implemented")
end
```

src/combinators/superpose.jl
```diff
@@ -76,13 +78,19 @@ oneplus(x::ULogarithmic) = exp(ULogarithmic, log1pexp(x.log))
     return dμ_dα / oneplus(dβ_dα) + dν_dβ / oneplus(dα_dβ)
 end
 
-using LogExpFunctions
+function density_def(s::SuperpositionMeasure, x)
+    error("Not implemented")
```
cscherrer, line 82:
```suggestion
    T = typeof(s)
    msg = """
    Not implemented: There is no method
    density_def(::$T, x)
    """
    error(msg)
```

src/combinators/smart-constructors.jl
```diff
@@ -131,5 +143,4 @@ function kleisli(::Type{M}; param_maps...) where {M}
     kleisli(M, nt)
 end
 
-
 kleisli(k::ParameterizedKleisli) = k
```
cscherrer, line 146:
```suggestion
kleisli(k::ParameterizedKleisli) = k

```

test/combinators/superpose.jl
```diff
@@ -0,0 +1,24 @@
...
+    μs = SuperpositionMeasure([μ, ν])
+    @test μs isa SuperpositionMeasure{<:AbstractVector{<:AbstractMeasure}}
+    @test_throws ErrorException density_def(μs, 0)
+    @test basemeasure(μs).components == SuperpositionMeasure([CountingMeasure(), CountingMeasure()]).components
+
+    μ2 = μ + μ
+    @test μ2 isa WeightedMeasure
+    @test μ2 == superpose(μ, μ)
+    @test basemeasure(μ2) == μ
+end
```
cscherrer, line 24:
```suggestion
end

```

cscherrer (changes requested):

PR #49, opened 2022-05-05
Dev

cscherrer:
## Kernel/Kleisli contention

We originally had `Kernel`. But this is very overloaded across the ecosystem. Also, what we have isn't quite a Markov kernel, since it maps to more general measures. We had changed it to `Kleisli`, which is accurate, but the term is not in wide wide use, especially among those doing stats work.

This PR tries to find a middle ground with `TransitionKernel`. It's descriptive and more intuitive, while still being accurate. The smart constructor has been changed back to `kernel`, but we're no longer exporting it in order to avoid conflicts with other packages.

## Conditional meausres

Soss and now Tilde use `|` for conditional models. But this should really be at the level of measures. So you can not express `measure | constraint` in the case where `measure` is over NamedTuples, and `constraint` is a particular named tuple.

One thing that may be surprising is that `|` in this PR does not change the dimensionality, but instead transforms the constrained dimension to a Dirac measure. I think this will be easier to later generalize to constraints that are not axis-aligned.

## Likelihoods

Some updates here, as well as introduction of `log_likelihood_ratio`. 

## Pointwise Products

Slot names are changed to `prior` and `likelihood` for easier interpretation. I've added an `iterate` method, allowing `μ, ℓ = (p::PointwiseProductMeasure)`.

## Power measures

`logdensity_def` is simplified to use `sum` instead of an explicit `for` loop. Hopefully this will later make it easier to port to GPU methods.

## PowerWeighted

In for example GLMs or Bayesian bootstrap, we want to consider each observation to have some weight. This is an exponent on the likelihood factor, or equivalently a multiplicative weight on the log-likelihood. Since `^` is already taken (and much more common than this) I've added
```julia
(d::AbstractMeasure) ↑ α = powerweightedmeasure(d, α)
```


cscherrer:
Hmm, tests passing on 1.8 and nightly, but not 1.6 or 1.7...

src/combinators/likelihood.jl
```diff
@@ -109,9 +115,9 @@ struct Likelihood{K,X} <: AbstractLikelihood
     k::K
     x::X
 
-    Likelihood(k::K, x::X) where {K<:AbstractKleisli,X} = new{K,X}(k,x)
+    Likelihood(k::K, x::X) where {K<:AbstractTransitionKernel,X} = new{K,X}(k,x)
     Likelihood(k::K, x::X) where {K<:Function,X} = new{K,X}(k,x)
-    Likelihood(μ, x) = Likelihood(kleisli(μ), x)
+    Likelihood(μ, x) = Likelihood(kernel(μ), x)
```
theogf, line 120: `kernel` takes two arguments no?

src/combinators/likelihood.jl
```diff
@@ -128,10 +134,37 @@ function Base.show(io::IO, ℓ::Likelihood)
...
-@inline function logdensity_def(ℓ::Likelihood, p)
-    return logdensity_def(ℓ.k((p,)), ℓ.x)
-end
+# basemeasure(ℓ::Likelihood, p) = basemeasure(ℓ.k(p), ℓ.x)
+
+# basemeasure(ℓ::Likelihood) = @error "Likelihood requires local base measure"
+
+export likelihood
+
+likelihood(k, x, ::NamedTuple{()}) = Likelihood(k, x)
```
theogf, line 147: It would be great to have some docs on how to use `likelihood`

src/combinators/likelihood.jl
```diff
@@ -128,10 +134,37 @@ function Base.show(io::IO, ℓ::Likelihood)
...
+parameters. This is computed as
+
+    logdensity_rel(ℓ.k(p), ℓ.k(q), ℓ.x)
+
+Since `logdensity_rel` can leave common base measure unevaluated, this can be
+more efficient than
+
+    logdensityof(ℓ.k(p), ℓ.x) - logdensityof(ℓ.k(q), ℓ.x)
+"""
+log_likelihood_ratio(ℓ::Likelihood, p, q) = logdensity_rel(ℓ.k(p), ℓ.k(q), ℓ.x)
```
theogf, line 168: Super nice! It should probably be exported no  (and tested)?

src/density.jl
```diff
@@ -69,7 +69,7 @@ function _densitymeasure(f, base, _)
 end
 
 @inline function insupport(d::DensityMeasure, x)
-    ifelse(insupport(d.base, x), logdensityof(d.f, x) > -Inf, false)
+    insupport(d.base, x) == true && logdensityof(d.f, x) > -Inf
```
theogf, line 72:
```suggestion
    insupport(d.base, x) && isfinite(logdensityof(d.f, x))
```

src/kernel.jl
```diff
@@ -0,0 +1,86 @@
...
+"""
+    kernel(f, M)
+    kernel((f1, f2, ...), M)
+
+A kernel `κ = kernel(f, m)` returns a wrapper around a function `f` giving the
+parameters for a measure of type `M`, such that `κ(x) = M(f(x)...)` respective
+`κ(x) = M(f1(x), f2(x), ...)`
+
+If the argument is a named tuple `(;a=f1, b=f1)`, `κ(x)` is defined as
+`M(;a=f(x),b=g(x))`.
```
theogf, line 23: That looks incorrect

src/combinators/conditional.jl
```diff
@@ -25,4 +25,16 @@ upcoming changes.
 """
 Base.:|(μ::AbstractMeasure, constraint) = ConditionalMeasure(μ, constraint)
```
theogf, line 26: Should there also be a smart constructor `condition`? `|` might be hard to find.

theogf (commented):
I did not look at everything in detail (this is a massive PR). But generally, it looks like many tests are missing, I think I caught some wrong implementations which would have been caught with some unit tests. At some point, I would be happy to help with the reorganization of the tests for the package.
I would not delete completely `kleisli`, it's still a nice thing to know and learn, and it can be exported. Maybe just leave it as an alias to `kernel` ? It's just my opinion though.

src/combinators/likelihood.jl
cscherrer: `kernel(::ParameterizedMeasure)` gives you a way to just replace the parameters.
```julia
julia> k = kernel(Normal(0,1))
f

julia> k((3,4))
Normal(μ = 3, σ = 4)
```

That `f` is pretty weird, it's because I return a function with multiple methods (so it can't be anonymous). Maybe there's a better way to do that, or I could give it a different name like `transitionkernel`

src/density.jl
cscherrer: We need the `== true` because `insupport(d.base, x)` might return `Static.True()`.

I used `> -Inf` because a result of `Inf` would mean we have singular measure, but we could still be in the support. OTOH by construction our `densityof` is never infinite, so maybe we could change this.

src/kernel.jl
cscherrer: Hmm, yeah I should review the docs. Good call.

src/combinators/conditional.jl
cscherrer: Yes, I like that.

src/combinators/likelihood.jl
theogf: Ah I see! But it only happens on `ParametrizedMeasure` then? this could be type-constrained in the function?

cscherrer: For now at least, I think that's right. Ok yeah, let's restrict the type.

src/kernel.jl
cscherrer: Let's have a separate docs PR after this:
https://github.com/cscherrer/MeasureBase.jl/issues/50

src/combinators/conditional.jl
cscherrer: added

src/combinators/likelihood.jl
cscherrer: Let's have a PR after this one to address docs
https://github.com/cscherrer/MeasureBase.jl/issues/50

PR #56, opened 2022-06-01
Standard-measures

cscherrer:


src/standard/stdnormal.jl
```diff
@@ -1,12 +1,11 @@
-struct StdNormal <: AbstractMeasure end
+struct StandardNormal <: AbstractMeasure end
 
-export StdNormal
+export StandardNormal
 
-insupport(d::StdNormal, x) = true
-insupport(d::StdNormal) = Returns(true)
+insupport(d::StandardNormal, x) = true
```
theogf: Generally, we always assume `x` is a scalar?

src/standard/stduniform.jl
```diff
@@ -1,22 +1,21 @@
...
+Base.rand(rng::Random.AbstractRNG, ::Type{T}, ::StandardUniform) where {T} = randn(rng, T)
 
-struct StdUniform <: AbstractMeasure end
+struct StandardUniform <: AbstractMeasure end
 
-export StdUniform
+export StandardUniform
 
-insupport(d::StdUniform, x) = zero(x) ≤ x ≤ one(x)
+insupport(d::StandardUniform, x) = zero(x) ≤ x ≤ one(x)
```
theogf: You are having the same code twice!

test/runtests.jl
```diff
@@ -37,6 +37,31 @@ test_measures = [
...
+    Dirac(0.0) + StdUniform()
+    SpikeMixture(StdUniform(), 0.2)
+
+    StdExponential()^3
+    StdExponential()^(2, 3)
+    3 * StdExponential()
+    StdExponential()
+    0.2 * StdExponential() + 0.8 * Dirac(0.0)
+    Dirac(0.0) + StdExponential()
+    SpikeMixture(StdExponential(), 0.2)
```
theogf, line 63: You probably need to update the names here too

src/standard/stduniform.jl
```diff
@@ -8,14 +8,3 @@ insupport(d::StdUniform, x) = 0 ≤ x ≤ 1
 @inline basemeasure(::StdUniform) = Lebesgue()
 
 Base.rand(rng::Random.AbstractRNG, ::Type{T}, ::StdUniform) where {T} = randn(rng, T)
-
-struct StdUniform <: AbstractMeasure end
-
-export StdUniform
-
-insupport(d::StdUniform, x) = zero(x) ≤ x ≤ one(x)
```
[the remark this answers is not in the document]
theogf: You kept `insupport(d::StdUniform, x) = 0 ≤ x ≤ 1`  instead of `insupport(d::StdUniform, x) = zero(x) ≤ x ≤ one(x)`

src/standard/stdnormal.jl
theogf: Maybe that's a stupid question but is there any check for it?

src/standard/stduniform.jl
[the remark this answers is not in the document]
cscherrer: I had thought this might help the compiler, but I'm surprised to see that's not the case. Say we define
```julia
f(x) = 0 ≤ x ≤ 1
g(x) = zero(x) ≤ x ≤ one(x)
```

These usually generate identical code. The one exception I've found so far is

```julia
julia> using Static

julia> @code_llvm f(static(0.0))
;  @ REPL[169]:1 within `f`
define i8 @julia_f_12050() #0 {
top:
  ret i8 1
}

julia> @code_llvm g(static(0.0))
;  @ REPL[170]:1 within `g`
; Function Attrs: noreturn
define nonnull {}* @julia_g_12052() #0 {
top:
  %0 = call nonnull {}* @"j_<=_12054"() #3
  call void @llvm.trap()
  unreachable
}
```

src/standard/stdnormal.jl
theogf: Ok I'll just leave it as an issue then.

theogf (approved):

PR #62, opened 2022-06-15
Add vartransform

oschulz:


src/effndof.jl
```diff
@@ -0,0 +1,33 @@
+"""
+    effndof(μ)
+
+Returns the effective number of degrees of freedom of variates of
```
cscherrer, line 4: Why "effective"? Doesn't DOF in the usual sense already account for this?

src/effndof.jl
```diff
@@ -0,0 +1,33 @@
...
+measure-like object `μ`.
```
cscherrer, line 5: "measure-like" is confusing, what does this mean?

src/effndof.jl
```diff
@@ -0,0 +1,33 @@
...
+    MeasureBase.require_same_effndof(a, b)::Nothing
+
+Check if `a` and `b` have the same effective number of degrees of freedom
+according to [`MeasureBase.effndof`](@ref).
+"""
+function require_same_effndof end
+
+ChainRulesCore.rrule(::typeof(require_same_effndof), a, b) = nothing, _nogradient_pullback2
+
+function require_same_effndof(a, b)
```
cscherrer, line 26: Can you tell me more about how these `require_*` functions are used?

src/insupport.jl
```diff
@@ -0,0 +1,44 @@
...
+function require_insupport(μ, x::AbstractArray{T,N}) where {T,N}
+    if !insupport(μ, x)
+        throw(ArgumentError("x is not within the support of μ"))
+    end
+    return nothing
+end
+
+
+"""
+    MeasureBase.check_varshape(μ, x)::Nothing
```
cscherrer, line 36: From the pattern above, should this be `require_varshape`?

src/vartransform.jl
```diff
@@ -0,0 +1,216 @@
...
+Indicates that no (default) pullback measure is available for measures of
+type `MU`.
+
+See [`MeasureBase.vartransform_origin`](@ref).
+"""
+struct NoTransformOrigin{MU} end
+
+
+"""
+    MeasureBase.vartransform_origin(μ)
```
cscherrer, line 13: What's the significance of "origin" in the name?

src/vartransform.jl
```diff
@@ -0,0 +1,216 @@
...
+
+Default measure to pullback to resp. pushforward from when transforming
```
cscherrer, line 15: I don't understand "to pullback to resp. pushforward"

src/vartransform.jl
```diff
@@ -0,0 +1,216 @@
...
+    struct MeasureBase.NoVarTransform{NU,MU} end
+
+Indicates that no transformation from a measure of type `MU` to a measure of
+type `NU` could be found.
+"""
+struct NoVarTransform{NU,MU} end
+
+
+"""
+    f = vartransform(ν, μ)::Function
```
cscherrer, line 53: I think `::Function` is to constraining. If you have functions $f:\mathbb{R}^m \to X$ and $g : \mathbb{R}^n \to Y$, you should be able to find an $h : \mathbb{R}^{m+n} \to X\times Y$. TransformVariables does this by internally passing around the "current index" for the transformation. This form composes well but it awkward to call. This is then wrapped in a function that says to start at one.

Without a mechanism like this, I think we're going to end up sacrificing performance or composability

src/vartransform.jl
```diff
@@ -0,0 +1,216 @@
...
+according to `ν`.
+
+The [pushforward measure](https://en.wikipedia.org/wiki/Pushforward_measure)
+from `μ` under `f` is is equivalent to `ν`.
+
+If terms of random values this implies that `f(rand(μ))` is equivalent to
+`rand(ν)` (if `rand(μ)` and `rand(ν)` are supported).
+
+The resulting function `f` should support
+`ChangesOfVariables.with_logabsdet_jacobian(f, x)` if mathematically well-defined,
```
cscherrer, line 71: Do we have a way to specify that some transforms should not use the `logabsdet_jacobian` correction? We've talked about discrete measures, but there's also `OrthoLebesgue`, which is defined as Lebesgue measure on an affine space. The "ortho" is because the logjac is ignored; the transform is only used to determine the support.

src/vartransform.jl
```diff
@@ -0,0 +1,216 @@
...
+
+"""
+    MeasureBase.select_vartransform_intermediate(a, b)
+
+Selects one of two candidate pullback measures `a, b` to use as an
+intermediate in variate transformations.
+
+See [`MeasureBase.vartransform_intermediate`](@ref).
+"""
+function select_vartransform_intermediate end
```
cscherrer, line 207: I don't understand this, could we add a small example?

cscherrer (commented):
Thanks @oschulz ! I have some questions, may have a better basis for review after I understand some things better

src/effndof.jl
oschulz: `require_same_effndof` is used in the `VarTransformation` constructor and the generic `vartransform` to fail early if things aren't compatible.

The `rrule` is there to remove the cost of that check from the pullback, of course.

src/vartransform.jl
oschulz: Ok, I'll remove it

src/insupport.jl
oschulz: I moved it to DistributionMeasures, it's not used in MB and would be a new API that measures would have to implement.

src/vartransform.jl
oschulz: I took the `::Function` out. The result of `vartransform(ν, μ)` should still be callable as `vartransform(ν, μ)(x)` though.

src/effndof.jl
cscherrer: `getdof`

oschulz: Called `getdof` now.

src/effndof.jl
oschulz: Called `check_dof` now.

src/standard/stdlogistic.jl
```diff
@@ -0,0 +1,21 @@
+struct StdLogistic <: StdMeasure end
+
+export StdLogistic
+
+@inline _isreal(x) = false
+@inline _isreal(x::Real) = true
```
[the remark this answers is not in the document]
oschulz: Ok, I took it back out.

src/combinators/power.jl
```diff
@@ -100,3 +100,6 @@ end
         dynamic(insupport(p, xj))
     end
 end
+
+
+@inline getdof(μ::PowerMeasure) = getdof(μ.parent) * prod(map(length, μ.axes))
```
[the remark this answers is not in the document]
oschulz: I think ` prod(map(length, μ.axes))` is much lighter on the compiler.

src/standard/stdlogistic.jl
```diff
@@ -0,0 +1,21 @@
...
+@inline basemeasure(::StdLogistic) = Lebesgue()
+
+@inline vartransform_def(::StdUniform, ::StdLogistic, x::Real) = logistic(x)
+@inline vartransform_def(::StdLogistic, ::StdUniform, x::Real) = logit(x)
+
+@inline getdof(::StdLogistic) = static(1)
+
+@inline Base.rand(rng::Random.AbstractRNG, ::Type{T}, ::StdLogistic) where {T} = logit(rand(rng, T))
+
+@inline StdMeasure(::typeof(randn)) = StdLogistic()
```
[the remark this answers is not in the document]
oschulz: Ah, that's a mistake!

src/combinators/power.jl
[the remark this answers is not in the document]
cscherrer: Strange if that's the case, but ok :)

src/vartransform.jl
oschulz: Do we stay with "origin"?

cscherrer:
Looking good! Can you put the examples so far into the tests?

src/vartransform.jl
oschulz: We can easily change it later without breaking much.

test/vartransform.jl
```diff
@@ -0,0 +1,44 @@
...
+            @test @inferred(with_logabsdet_jacobian(f, x)) isa Tuple{supertype(y),Real}
+            @test @inferred(with_logabsdet_jacobian(inverse(f), y)) isa Tuple{supertype(x),Real}
+            y2, ladj_fwd = with_logabsdet_jacobian(f, x)
+            x2, ladj_inv = with_logabsdet_jacobian(inverse(f), y)
+            @test x ≈ x2
+            @test y ≈ y2
+            @test ladj_fwd ≈ - ladj_inv
+            @test ladj_fwd ≈ logdensityof(μ, x) - logdensityof(ν, y)
+        end
+    end
```
cscherrer, line 29: Should this or something similar go in `MeasureBase.Interface`?

src/getdof.jl
```diff
@@ -0,0 +1,75 @@
+"""
+MeasureBase.NoDOF{MU}
```
mschauer, line 2:
```suggestion
    MeasureBase.NoDOF{MU}
```

test/vartransform.jl
oschulz: Would that break tests for measures that are not vartransform-compatible (yet)?

mschauer:
Can you add one or two paragraphs of high-level documentation in a separate `.md` in https://github.com/cscherrer/MeasureTheory.jl/tree/master/docs/src ?

test/vartransform.jl
cscherrer: I'd have it as a separate test so it can be opt-in, at least for now

oschulz:
> Can you add one or two paragraphs of high-level documentation in a separate .md in https://github.com/cscherrer/MeasureTheory.jl/tree/master/docs/src ?

Will do!

test/vartransform.jl
oschulz: Done - I'll improve it a bit over time. A future addition would be to add an (optional) `getjacobian` argument, similar to what we did in [ChangesOfVariables.test_with_logabsdet_jacobian](https://github.com/JuliaMath/ChangesOfVariables.jl/blob/acd2a0c8d59bf664d4c3c2404f4b58ac26f52647/src/test.jl#L24), to compare with a calculated LADJ.

oschulz:
> > Can you add one or two paragraphs of high-level documentation in a separate .md in https://github.com/cscherrer/MeasureTheory.jl/tree/master/docs/src ?
> 
> Will do!

Here's first draft: https://github.com/cscherrer/MeasureTheory.jl/pull/210

Obviously will need to wait before the new API is released on MB. :-)

src/getdof.jl
```diff
@@ -0,0 +1,75 @@
...
+"""
+    MeasureBase.checked_var(μ::MU, x::T)::T
+
+Return `x` if `x` is a valid variate of `μ`, throw an `ArgumentError` if not,
+return `NoVarCheck{MU,T}()` if not check can be performed.
+"""
+function checked_var end
+
+# Prevent infinite recursion:
+@propagate_inbounds _default_checked_var(::Type{MU}, ::MU, ::Any) where MU = NoVarCheck{MU,T}
```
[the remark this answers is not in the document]
oschulz: Thanks, indeed! Fixex.

src/combinators/transformedmeasure.jl
```diff
@@ -13,3 +13,99 @@ function params(::AbstractTransformedMeasure) end
 function paramnames(::AbstractTransformedMeasure) end
 
 function parent(::AbstractTransformedMeasure) end
+
+
+abstract type TransformVolCorr end
```
cscherrer, line 18: Do you have a reason to use `VolCorr` instead of `LogJac`? That's more familiar I think

oschulz: `VolCorr` emphasizes the purpose over the mechanism `LogJac`, but I have no strong opinion on it.

cscherrer:
@mschauer I haven't used this yet in detail, but since @oschulz has had success with the approach and has included lots of tests for the new code I think it's good to go. Want to give you a chance to look before a merge and release, but I'd like to do that today if possible so we're not holding up DistributionMeasures updates.

cscherrer (approved):

cscherrer:
Also, I ran `JET.report_package` on this and it's looking good

oschulz:
Ok, now we can transform Dirac as well.

mschauer (commented):
I am somewhat unhappy with calling it vartransform. For me var in this context evokes “variance” More importantly, once you have that map, it does the same as “as(post)”, but the interface seems to be different (one gives a map, one gives an input to transform)

oschulz:
> I am somewhat unhappy with calling it vartransform. For me var in this context evokes “variance”

Yes, I see your point. We should converge on a name quickly, though, since I have severl things "in flight" that depend in it - do you have a preference?

> More importantly, once you have that map, it does the same as “as(post)”

What is `post` here, resp. what does this imply?

cscherrer:
> What is `post` here, resp. what does this imply?

In Soss and Tilde, we often use this name for a posterior. So `as(post)` gives a transform from $\mathbb{R}^n$ to the support of the posterior (the same as that of the prior).

`as` is currently in MeasureTheory (not MeasureBase) because it works in terms of TransformVariables.jl. `vartransform` is conceptually similar, with the big difference being that it works in terms of the measure itself, while `as` only depends directly on the support of a measure.

I think we can build toward a common interface so these can be a single function, then the naming issues will clear up. It would be great if this could be community-wide, though I think it's important to prioritize functionality over backward-compatibility (or MT wouldn't exist in the first place).

@oschulz are you willing to commit to a path like this? It will require internal name changes in BAT down the line, though I think you could still define `vartransform` and make that available to users when we get to that point.

@mschauer, if we agree we're moving toward a common interface and `vartransform` is a short-term placeholder, does that address your concerns?

oschulz:
> In Soss and Tilde, we often use this name for a posterior. So as(post) gives a transform from  to the support of the posterior (the same as that of the prior).

Ah, yes. I would say that `vartransform` is really different - you will typically use it with the prior, not the posterior, and of course it does exact transformations between measures, not just to/from an unconstrained flat space.

`as` only nees/uses information on the support, `vartransform` couldn't operate directly on a posterior, at least not in a mechanistic/automatic fashion.

> as is currently in MeasureTheory (not MeasureBase) because it works in terms of TransformVariables.jl. vartransform is conceptually similar, with the big difference being that it works in terms of the measure itself, while as only depends directly on the support of a measure.

Exactly, I think `as` has quite as different focus, in addition to a different API structure.


> I think we can build toward a common interface so these can be a single function

I really don't want MeasureBase to depend on TransformVariables, it's quite heavy. Also, having something that generates and actual callable with native support for InverseFunctions and ChangesOfVariables. is imporant for my applications.

What I could see is using the approach and API developed in this PR, and expressing the unconstrained transformation simply as `vartransform(Unconstrained(mu), mu)` like we discussed offline. That could then, for example use TransformVariables (and maybe optionally Bijectors) internally.

So I think `as` can be expressed in terms of `vartransform`, but not the other way rounds. But I don't think what `as` currently does and how it's used can be merged with `vartransform` under a single function name.


> It would be great if this could be community-wide, though I think it's important to prioritize functionality over backward-compatibility (or MT wouldn't exist in the first place).

Absolutely - but with MeasureBase as the central depencency, right? `vartransform` (and it's default implementation parts) does need quite a few MB concepts.



> @oschulz are you willing to commit to a path like this? It will require internal name changes in BAT down the line

The names in BAT are currently different anyway, a full rename is not a problem. But the approach, having a function `generate_a_varitate_transform(to, from)` that returns a callable on variates is very important to me. And the way it currently works (how transforms are "declared") is not ad-hoc, and tested in real-world applications. We can certainly make changes there if we come up with a better way, but for now I'd like to get it into MB if at all possible.

As far as function names are concerned, I fully open to switch to another names now, if we can select then quickly.

cscherrer:
> Ah, yes. I would say that `vartransform` is really different - you will typically use it with the prior, not the posterior, and of course it does exact transformations between measures, not just to/from an unconstrained flat space.

But the result is a very similar kind of thing, even if how it gets there is very different.
 
> `as` only needs/uses information on the support, vartransform couldn't operate directly on a posterior, at least not in a mechanistic/automatic fashion (needs a normalizing flow optimizer or similar).

Right, so calling on a posterior would return `NoTransform()` or similar.

> I really don't want MeasureBase to depend on TransformVariables, it's quite heavy. Also, having something that generates and actual callable with native support for InverseFunctions and ChangesOfVariables. is imporant for my applications.

Agreed. I want to take TV as inspiration. If it can come to literally use the same interface that's great, but we also have concrete needs for we need to be sure are addressed. And I agree transforms should be callable on elements of the support.

> The names in BAT are currently different anyway, a full rename (now or later) is not a problem. But the approach, having a function generate_a_varitate_transform(to, from) that returns a callable on variates is very important to me. And the way it currently works (how transforms are "declared") is not ad-hoc, and tested in real-world applications. We can certainly make changes there if we come up with a better way, but for now I'd like to get it into MB if at all possible.

Yes, as I see it (@mschauer correct me if I'm missing the mark) concerns are (1) a confusing name, and (2) having transforms conceptually under one umbrella, even if they're determined very differently. 

I think (1) can be addressed easily - just a matter of coming up with an alternative name - and (2) will be a longer term goal, maybe a few months down the line.

oschulz:
> Right, so calling on a posterior would return NoTransform() or similar.

Oh, sure, it should be legal to call it on anything that supports the measure interface. And for some special posteriors (e.g. a posterior using a likelihood with conjugate prior) a transform could of course be available.

> concerns are (1) a confusing name

Yes, I didn't notice before, but I think @mschauer  is right, `vartransform` could be misunderstood. How about any of these

* `transformation_to(to, from)`
* `variate_transform(to, from)`
* `measureable_function(to, from)`
* `transform_function(to, from)`
* `measure_transform(to, from)`

> and (2) will be a longer term goal, maybe a few months down the line.

Sounds good. I think we''' certainly be able to find way to bring "generic unconstrained" transforms under this umbrella as well (e.g. via `Unconstrained(mu) <: AbstractMeasure` and/or API extensions).

oschulz:
I proposed some possible names above. I like `transformation_to` - it's obvious in it's intent, I think. But anything clear is fine with me.

mschauer:
One could also derive a name from the view that 
the function returns a map that transports the measure nu into the measure mu (transport is not as overused as transform). Otherwise I’d say “transform_map”

oschulz:
> Otherwise I’d say “transformation_map” or as you propose, transformation_to

I'd suggest `transformation_to`, then - `transformation_map` sounds a bit redundant between "transformation" and "map". 

`transform_to` also clearly implies the argument order (which matches the "first target then source" that we also have in `Base.convert` and other places. It is also the natural choice, since if that if we add mutable transforms at some point we'll want to have measures and their values next to each other, e.g. in  `transform_def!(y, trg, src, x)`.)

oschulz:
If we want to stress the transport aspect, I'd suggest `transport_map(nu, mu)` (with `transport_def(nu, mu, x)` for the internal function, and `nu` the target and `mu` the source).

oschulz:
Ok, new list for bikeshedding:

* `transformation_to(nu, mu)`
* `transformation_map(nu, mu)`
* `transport_to(nu, mu)`
* `transport_map(nu, mu)`

I like the ones with `_to` better, but I'm Ok with all of them.

@cscherrer ,  @mschauer , any picks?

cscherrer:
I think it's important to try to distinguish between a _measure_ transformation and a _support_ transformation (that can be pushed forward or pulled back for some measure). And the to/from thing could be confusing.

If the concept of a transport is meaningful when the spaces are different, I like `transport_to`

oschulz:
> I think it's important to try to distinguish between a _measure_ transformation and a _support_ transformation (that can be pushed forward or pulled back for some measure). And the to/from thing could be confusing.
> 
> If the concept of a transport is meaningful when the spaces are different, I like `transport_to`

If I read https://en.m.wikipedia.org/wiki/Transportation_theory_(mathematics) correctly, the spaces can be different. Correct, @mschauer?

We do not guarantee that we return an *optimal* transport, though (we shovel the earth correctly but not necessarily with minimum effort).

mschauer:
Exactly, we also don’t promise an optimal transport, right? :-)

oschulz:
> Exactly, we also don’t promise an optimal transport, right? :-)

Indeed no, just a "correct transport, not an optimal one.

cscherrer:
@mschauer if we change to `transport_to` do you think we're good to go?

I'm going to be out for a while, so once you're both happy with this please go ahead and merge, and tag a release

mschauer (approved):

mschauer:
I should not tag releases on the phone but I am happy with the PR

oschulz:
> I should not tag releases on the phone but I am happy with the PR

Ok, let me do the rename, then we'll tag?

oschulz:
Ok, renamed, let's see if CI goes through.

@mschauer Ok like this?

oschulz:
Ah, stop, need to rename `vartransform_origin` to

oschulz:
Ok rename complete.

oschulz:
> I should not tag releases on the phone but I am happy with the PR

I can do the squash-merge, increase version number and tag, if there are no objections?

oschulz:
This will be v0.11.0, not v0.10.0, right?

oschulz:
> This will be v0.11.0, not v0.10.1, right?

Ah, yes - has to be, since it moves out StdNormal.

oschulz:
Released. :-)

oschulz:
I apologize, I overlooked one name - we still have the `VarTransform` type. Should I make a v0.12 (we could name it `TransportTo`? It's not a type users will interact with much, but maybe best to fix that naming inconsistency now instead of later?

oschulz:
I did a grep for var ... we also forgot about `test_vartransform`, `NoTransport`, `checked_var`, `NoVarCheck` and a few internal functions during the rename (oops).

oschulz:
#63 should fix it - grep shows no remaining "...var..." where it doesn't belong.

PR #63, opened 2022-06-20
Address some transform-related naming inconsistencies

oschulz:
We forgot a few additional types and functions when we renamed `vartransform` to `transport_to` during the course of #62. This renames them to make things more consistent:

* Rename VarTransformation to TransportFunction
* Rename test_vartransform to test_transport
* Rename NoVarTransform to NoTransport
* Rename checked_var to checked_arg
* Rename NoVarCheck to NoArgCheck
* Rename internal transport helper functions

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "MeasureBase"
 uuid = "fa1605e6-acd5-459c-a1e6-7e635759db14"
 authors = ["Chad Scherrer <chad.scherrer@gmail.com> and contributors"]
-version = "0.11.0"
+version = "0.12.0"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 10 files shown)


oschulz:
Most names changes are straightforward, I think. I'm not sure about `checked_var` - I rename it to `checked_arg` here, but maybe that could also be misunderstood? Would `checked_variate`  be better?

mschauer:
I am happy with `checked_arg`

mschauer (approved):

oschulz:
> If it's not very user facing, how about `check_support_type`?

It doesn't just check the type, it can also check the size of arrays, etc.

oschulz:
Are you happy with the state of this @cscherrer

cscherrer (approved):

cscherrer:
@oschulz is this a minor or patch release?

oschulz:
> @oschulz is this a minor or patch release?

It'll have to be v0.12, some of these things have docstrings. So probably best to fix these naming inconsistencies now, I'd say.

oschulz:
Ok, if I go ahead and tag this v0.12 and release it?

PR #65, opened 2022-06-20
working on fixing Aqua tests

cscherrer:


oschulz:
> First and most importantly (because it can impact performance), it fixes some unbound arguments found by Aqua

Oh nice - yes, those can indeed affect performance.

cscherrer:
> Regarding `MeasureBase.NoDOF{MU}` and the like I had in docstrings - I put them in there to indicate to the reader that those functionalities are official but not exported.

I see, that's a good idea. I'll roll those back

cscherrer:
@oschulz do you think there's anything to do with `ambiguities.jl`?

PR #67, opened 2022-06-23
Adapt to Static.jl v0.7

oschulz:


cscherrer:
We need to be careful with this one:
https://github.com/SciML/Static.jl/issues/66

oschulz:
Indeed.

Issue #70, opened 2022-07-01
`latentof` and `jointof`


cscherrer:
Say we have a Soss or Tilde model like
```julia
cauchy = @model begin
    a ~ Normal()
    b ~ Normal()
    return a / b
end
```
Models like this consist of a latent space (named tuples over `a` and `b`) and a pushforward (taking a named tuple `nt` to `nt.a / nt.b`). The end result is a measure over $\mathbb{R}$. But for inference, we often need to work not with the result of the pushforward, but with the latent space. This leaves us needing a way to construct this latent measure.

@oschulz and I discussed this today, and realized we can add a `latentof` function that takes a measure and returns a measure over its latent space. For measures `m` not expressed in terms of a pushforward, we'll have `latentof(m) == m`.

It's also often convenient to express the joint measure, taking values on the product of the latent value and return value. With the function
```julia
diagonal_pair(x) = x => x
```
we can have the default
```julia
jointof(m::AbstractMeasure) = pushfwd(diagonal_pair, m)
```
For `cauchy` above, an element of the joint measure will look like `(a = 0.5, b = 0.4) => 1.2`.

One possible concern is that "joint" usually indicates a measure with elements in a product space, and not such a specific case as pairing a point in latent space with the corresponding pushforward. But this use of "joint" is unlikely to be useful in MeasureTheory.jl, and I haven't yet heard a better name for what we're trying to describe.

PR #72, opened 2022-07-11
latent etc

cscherrer:
Fixes https://github.com/cscherrer/MeasureBase.jl/issues/70

### DIFF
--- a/src/latent-joint.jl
+++ b/src/latent-joint.jl
@@ -1,5 +1,14 @@
-diagonal_pair(x) = x => x
+latentof(m) = m
+manifestof(m) = m
 
-latentof(m::AbstractMeasure) = m
-manifestof(m::AbstractMeasure) = m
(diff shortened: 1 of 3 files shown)


src/latent-joint.jl
```diff
@@ -0,0 +1,62 @@
...
+
+"""
+    jointof(m)
+
+$_LATENT_DOCSTRING
+"""
+function jointof(m)
+    fwd(x) = x => x
+    
+    function back(p::Pair)
```
oschulz, line 55: Isn't `back` a bit too generic?

cscherrer: Yes, but it's only in scope in a very limited context

oschulz: Sure, why not then.

oschulz:
Fine with me!

PR #78, opened 2022-08-04
likelihoodof

cscherrer:
This PR makes two kinds of changes:
1. Changes `likelihood` to `likelihoodof`, to allow the former to be more easily used as a local variable
2. Makes `Likelihood`s callable, by adding this method
```julia
(lik::AbstractLikelihood)(p) = exp(ULogarithmic, logdensityof(lik.k(p), lik.x))
```

In this way, computations are in terms of log-likelihood, but the result is in linear space with no additional overhead. This makes it easy to choose either.

There are a couple of reasons I like the approach in (2). First, it makes likelihoods easier to work with. It's natural and very standard to express a likelihood as a function. Calling `logdensityof` on a likelihood is not only more awkward, but it also can be confusing, leading people to think a likelihood is a density between two existing measures. Though we can think of a likelihood as a "proto-density", the subtlety in this is easily lost. 

Second, since we're calling these "likelihoods" and not "log-likelihoods", it's more clear that the result can be expected to be in linear space.

Note that we'll still keep the `logdensityof` method for likelihoods.

### DIFF
--- a/src/combinators/likelihood.jl
+++ b/src/combinators/likelihood.jl
@@ -121,6 +121,8 @@ struct Likelihood{K,X} <: AbstractLikelihood
     Likelihood(μ, x) = Likelihood(kernel(μ), x)
 end
 
+(lik::AbstractLikelihood)(p) = exp(ULogarithmic, logdensityof(lik.k(p), lik.x))
+
 DensityInterface.DensityKind(::AbstractLikelihood) = IsDensity()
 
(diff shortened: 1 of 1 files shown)


oschulz (approved):

oschulz:
LGTM, as long as we continue to support `logdensityof` for `Likelihood` long-term.

PR #83, opened 2022-09-02
explicit type parameters for NamedTuples

cscherrer:
This PR makes two small changes. 

First, the `::NamedTuple` method for `ParameterizedMeasure`s is changed to `::NamedTuple{K,T}`, which hints to the compiler that methods should be specialized.

Second, the method
```julia
(::Type{P})(; kwargs...) where {P<:ParameterizedMeasure}
```
is redundant, since this is already handled by KeywordCalls. This PR removes this method.

### DIFF
--- a/src/parameterized.jl
+++ b/src/parameterized.jl
@@ -60,8 +60,6 @@ function (::Type{P})(args...) where {N,P<:ParameterizedMeasure{N}}
     return C(NamedTuple{N}(args))::C{N,typeof(args)}
 end
 
-# (::Type{P})(; kwargs...) where {P<:ParameterizedMeasure} = P(NamedTuple(kwargs))
-
 function ConstructionBase.setproperties(
     d::P,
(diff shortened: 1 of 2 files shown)


src/parameterized.jl
```diff
@@ -60,7 +60,7 @@ function (::Type{P})(args...) where {N,P<:ParameterizedMeasure{N}}
     return C(NamedTuple{N}(args))::C{N,typeof(args)}
 end
 
-(::Type{P})(; kwargs...) where {P<:ParameterizedMeasure} = P(NamedTuple(kwargs))
+# (::Type{P})(; kwargs...) where {P<:ParameterizedMeasure} = P(NamedTuple(kwargs))
```
cscherrer, line 64:
```suggestion
```

oschulz (approved):

oschulz:
LGTM

PR #85, opened 2022-09-02
callable densities

cscherrer:
Fixes https://github.com/cscherrer/MeasureBase.jl/issues/91

This PR also reorganizes the code a little, adding `density-core.jl` and moving lots of functions there. This is because these functions are very different from `(Log)Density` and `(Log)DensityMeasure`.


cscherrer:
@oschulz what do you think of the dispatching approach here? Previouslywe had
```julia
∫(f::Function, base::AbstractMeasure) = DensityMeasure(funcdensity(f), base)

∫(f, base::AbstractMeasure) = _densitymeasure(f, base, DensityKind(f))
```

But this seems to assume `f` is either `<:Function` or `IsDensity`. In practice there are lots of callables that aren't either; I think dispatching on `Function` like this is pretty dangerous, and we should avoid it.

One option I considered was to just always call `funcdensity` or `logfuncdensity`. But there seems to be no `funcdensity(::FuncDensity)` for example (should there be?) , so currently we need to consider those cases separately.

oschulz:
> what do you think of the dispatching approach here? ... I think dispatching on Function like this is pretty dangerous

I agree. Can't we just use `DensityInterface.DensityKind` to do a trait-based dispatch? So just have `∫(f, base::AbstractMeasure)`, then then dispatch again in a helper function based on whether `DenstiyKind(f)` is a `IsDensity` (so it's a "DensityInterface-density"), or `NoDensity` (then we assume it's a "simple" density function) or a `HasDensity` (so a measure or similar, this should result in an error).

This is the kind of situation I had in mind when advocating for the `DensityKind` functionality in DensityInterface.

cscherrer:
I was a little concerned with mixing manual and "magic" versions of choosing log vs linear. So currently it's
```julia
∫(f, base) = DensityMeasure(funcdensity(f), base)

∫(f::FuncDensity, base) = DensityMeasure(f, base)

function ∫(f::LogFuncDensity, base)
    @error "Can't call `∫` on a `LogFuncDensity`; use `∫exp` instead"
end
```

If it's just `IsDensity`, how would we know whether the callable version should be in log space?

oschulz:
> If it's just IsDensity, how would we know whether the callable version should be in log space?

I would say we just use log space in that case.

cscherrer:
> I would say we just use log space in that case.

I think we're talking about different things. `∫` always creates a `DensityMeasure`, and `∫exp` always creates a `LogDensityMeasure`.  But I could change this code

```julia
∫(f, base) = DensityMeasure(funcdensity(f), base)

∫(f::FuncDensity, base) = DensityMeasure(f, base)

function ∫(f::LogFuncDensity, base)
    @error "Can't call `∫` on a `LogFuncDensity`; use `∫exp` instead"
end
```

to
```julia
∫(f, base) = _densitymeasure(f, base, DensityKind(f))

_densitymeasure(f, base, ::IsDensity) = DensityMeasure(f, base)
_densitymeasure(f, base, ::HasDensity) = @error "..."
_densitymeasure(f, base, ::NoDensity) = DensityMeasure(funcdensity(f), base)
```

if you like that better

cscherrer:
Now I'm second-guessing this... From an end-user's perspective, what's the difference between a `DensityMeasure` and `LogDensityMeasure`? Maybe we only need one type here, with two different constructors?

I think I'll try that, might simplify things. So we'll still have `Density` and `LogDensity`, because those act differently as callables.

src/density.jl
```diff
@@ -21,257 +30,129 @@ struct Density{M,B} <: AbstractDensity
...
+Users should not call `DensityMeasure` directly, but should instead call `∫(f,
+base)` (if `f` is a density) or `∫exp(f, base)` (if `f` is a log-density).
 """
-    unsafe_logdensityof(m, x)
-
-Compute the log-density of the measure `m` at `x` relative to `rootmeasure(m)`.
-This is "unsafe" because it does not check `insupport(m, x)`.
+struct DensityMeasure{F,B} <: AbstractMeasure
+    f::F
+    base::B
```
cscherrer, line 110: For a `mu::DensityMeasure`, there's nothing to prevent users from getting `mu.f`, which might be a density or log-density. How dangerous is this?

src/density.jl
```diff
@@ -21,257 +30,129 @@ struct Density{M,B} <: AbstractDensity
...
-    inμ || return convert(T, ifelse(inν, -Inf, NaN))
-    inν || return convert(T, Inf)
-
-    return unsafe_logdensity_rel(μ, ν, x)
-end
+∫(f, base) = _densitymeasure(f, base, DensityKind(f))
 
-"""
-    unsafe_logdensity_rel(m1, m2, x)
+_densitymeasure(f, base, ::IsDensity) = DensityMeasure(f, base)
```
cscherrer, line 137: In this implementation, users could call `∫(f, base)` or `∫exp(f, base)` and get the same result, if `DensityKind(f)==IsDensity()`.

oschulz:
We shouldn't dispatch on `LogFuncDensity` - that's just a convenience in DensityInterface, people can implement densities that have a completely different type (but have `DensityKind(d) === IsDensity()`).

oschulz:
Question, do we need to have separate functions `logdensity_rel(μ, base, x)` and `logdensity(μ, base)`, or can we just have `logdensity_rel(μ, base, x)` and `logdensity_rel(f.μ, f.base)` (similar to `logdensityof(μ, x)` and `logdensity(μ)` in DensityInterface)?

src/density.jl
oschulz: If they're callable, I'd say it's dangerous and too magical.

src/density.jl
oschulz: I don't think that should be allowed. I'd say `∫exp(f, base)` should only accept functions, not DensityInterface "density objects".

src/density.jl
cscherrer: Thanks, I was kind of leaning that way too. So maybe there's a way around this. If not, we could also muck with `getproperty`

cscherrer:
> We shouldn't dispatch on `LogFuncDensity` - that's just a convenience in DensityInterface, people can implement densities that have a completely different type (but have `DensityKind(d) === IsDensity()`).

Thanks, that makes sense



> Question, do we need to have separate functions `logdensity_rel(μ, base, x)` and `logdensity(μ, base)`, or can we just have `logdensity_rel(μ, base, x)` and `logdensity_rel(f.μ, f.base)` (similar to `logdensityof(μ, x)` and `logdensity(μ)` in DensityInterface)?

Great idea! I'll change to this

src/density.jl
oschulz: How about this?

```julia
struct DensityMeasure{D,B} <: AbstractMeasure
    d::D
    base::B
end

∫(density, measure) = _∫(DensityKind(density), density, measure) 
_∫(::IsDensity, density, measure) = DensityMeasure(density, base)
_∫(::HasDensity, density, measure) = throw(ArgumentError("This makes no sense"))
_∫(::NoDensity, density, measure) = DensityMeasure(funcdensity(density), base)

∫exp(logdensity, measure) = DensityMeasure(logfuncdensity(logdensity), measure)
```

So we'd always store a DensityInterface-compatible object in `DensityMeasure`.

cscherrer: Yes, I currently have
```julia
_densitymeasure(f, base, ::IsDensity) = DensityMeasure(f, base)
_densitymeasure(f, base, ::HasDensity) = @error "..."
_densitymeasure(f, base, ::NoDensity) = DensityMeasure(funcdensity(f), base)
```
and the struct looks like
```julia
struct DensityMeasure{F,B} <: AbstractMeasure
    f::F
    base::B

    function DensityMeasure(f::F, base::B) where {F,B}
        @assert DensityKind(f) == IsDensity()
        new{F,B}(f, base)
    end
end
```
I think construction is ok (but you should have a look anyway) - but destruction is a problem.

Maybe something like
```julia

function Base.propertynames(::DensityMeasure)
    (:density, :logdensity, :base)
end

function Base.getproperty(μ::DensityMeasure, s)
    if s == :density
        ....
    else if s == :logdensity
        ....
    else if s == :base
        return :getfield(μ, :base)
    end
end
```
?

cscherrer: Hmm, what to return is very tricky, because DensityInterface doesn't keep a representation of whether it's in log space

src/density.jl
cscherrer: What about `∫(f, base)`?

cscherrer: Maybe this does it?
```julia
function Base.propertynames(::DensityMeasure)
    (:density, :logdensity, :base)
end

function Base.getproperty(μ::DensityMeasure, s)
    if s == :density
        x -> densityof(getfield(μ, :f), x)
    else if s == :logdensity
        x -> logdensityof(getfield(μ, :f), x)
    else if s == :base
        return :getfield(μ, :base)
    end
end
```

src/density.jl
cscherrer: Maybe this does it?
```julia
function Base.propertynames(::DensityMeasure)
    (:density, :logdensity, :base)
end

function Base.getproperty(μ::DensityMeasure, s)
    if s == :density
        x -> densityof(getfield(μ, :f), x)
    else if s == :logdensity
        x -> logdensityof(getfield(μ, :f), x)
    else if s == :base
        return :getfield(μ, :base)
    end
end
```

oschulz: But do we need to destruct that way at all? In the end, people have to ask for `density_rel` and `logdensity_rel` explicitly anyway, right? If `DensityMeasure` stores DensityInterface-objects, isn't it enough to have

```julia
density_def(m::DensityMeasure, x) = densityof(m.d, x)
logdensity_def(m::DensityMeasure, x) = logdensityof(m.d, x)
```

so that `density_rel` and `logdensity_rel` will do the right thing?

oschulz: Access to the density and base part of `m` would then be the job of functions like `likelihoodof` and `priorof`, though we should find less Bayes-specific names.

oschulz: To be honest, I really don't like the "virtual property" approach here, it seems hacky (and might cause trouble with autodiff and similar too).

cscherrer: Right, but if we have a struct, people are very likely to use dot syntax to deconstruct it. Options I see for dealing with this are
1. Just let them, and correct them when they tell us it's broken
2. Make it very hard to do, e.g. `getproperty` could just error
3. Make `getproperty` available, and have it do something helpful

oschulz: > Right, but if we have a struct, people are very likely to use dot syntax to deconstruct it.

Sure, but then they get non-callable density objects, so those can't be "abused".

src/density.jl
oschulz: `∫(f, base)` would accept both density objects (like a `Likelihood`) and simple numeric non-log density functions.

cscherrer:
Ready for another look @oschulz :)

oschulz:
After these changes, what's the function now to get a "density object" compatible with DensityInterface?

oschulz:
LGTM

oschulz (approved):

Issue #87, opened 2022-09-09
Bring back StdNormal?


cscherrer:
We had `StdNormal` for a while, but dropped in in favor of `StdLogistic` because a Normal cdf requires `erf` or `erfc`, which are in SpecialFunctions.jl, a heavier dependency than we wanted for MeasureBase.

But the [Wikipedia page for normal distributions](https://en.wikipedia.org/wiki/Normal_distribution#Numerical_approximations_for_the_normal_CDF_and_normal_quantile_function) links to [this article](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.559.1540&rep=rep1&type=pdf), which has a fairly short algorithm for this:
```
Function Cumnorm(x As Double) As Double
    XAbs = Abs(x)
    If XAbs > 37 Then
        Cumnorm = 0
    Else
        Exponential = Exp(-XAbs ^ 2 / 2)
        If XAbs < 7.07106781186547 Then
            build = 3.52624965998911E-02 * XAbs + 0.700383064443688
            build = build * XAbs + 6.37396220353165
            build = build * XAbs + 33.912866078383
            build = build * XAbs + 112.079291497871
            build = build * XAbs + 221.213596169931
            build = build * XAbs + 220.206867912376
            Cumnorm = Exponential * build
            build = 8.83883476483184E-02 * XAbs + 1.75566716318264
            build = build * XAbs + 16.064177579207
            build = build * XAbs + 86.7807322029461
            build = build * XAbs + 296.564248779674
            build = build * XAbs + 637.333633378831
            build = build * XAbs + 793.826512519948
            build = build * XAbs + 440.413735824752
            Cumnorm = Cumnorm / build
        Else
            build = XAbs + 0.65
            build = XAbs + 4 / build
            build = XAbs + 3 / build
            build = XAbs + 2 / build
            build = XAbs + 1 / build
            Cumnorm = Exponential / build / 2.506628274631
        End If
    End If
    If x > 0 Then Cumnorm = 1 - Cumnorm
End Function
```

Maybe we should consider including this?

cscherrer:
In Julia this is
```julia
function Φ(x::T) where {T}
    xabs = abs(x)
    if xabs > 37 
        result = zero(T)
    else
        bell = exp(-xabs ^ 2 / 2)
        if xabs < 7.07106781186547 
            build = muladd(3.52624965998911E-02, xabs, 0.700383064443688)
[...]

Issue #86, opened 2022-09-08
Add `FixedRNG`


cscherrer:
I've been using this in Tilde lately:
```julia
struct FixedRNG <: AbstractRNG end

Base.rand(::FixedRNG) = one(Float64) / 2
Random.randn(::FixedRNG) = zero(Float64)
Random.randexp(::FixedRNG) = one(Float64)

Base.rand(::FixedRNG, ::Type{T}) where {T<:Real} = one(T) / 2
Random.randn(::FixedRNG, ::Type{T}) where {T<:Real} = zero(T)
Random.randexp(::FixedRNG, ::Type{T}) where {T<:Real} = one(T)

# We need concrete type parameters to avoid amiguity for these cases
for T in [Float16, Float32, Float64]
    @eval begin
        Base.rand(::FixedRNG, ::Type{$T}) = one($T) / 2
        Random.randn(::FixedRNG, ::Type{$T}) = zero($T)
        Random.randexp(::FixedRNG, ::Type{$T}) = one($T)
    end
end
```

I think we should move this to MeasureBase, together with a fallback method
```julia
testvalue(d) = rand(FixedRNG(), d)
```

In this way, any measure with a `rand` method will get a `testvalue` method for free.

Issue #82, opened 2022-08-22
Mass interface


cscherrer:
We need a `massof` and `logmassof` for measures.

Some discussion here:
https://github.com/JuliaMath/DensityInterface.jl/issues/8

cc @oschulz

PR #88, opened 2022-09-12
Dev

cscherrer:
- Add `FixedRNG` from Tilde (fixes https://github.com/cscherrer/MeasureBase.jl/issues/86)
- Add `StdNormal` from DistributionMeasures (fixes https://github.com/cscherrer/MeasureBase.jl/issues/87)
- Transport methods for pushforwards
- Rename `NoTransformOrigin` to `NoTransportOrigin`
- `testvalue(::Type{T}, m)`: optional `T` argument is new, works as in `rand`
- Rename `LebesgueMeasure` => `LebesgueBase`, `CountingMeasure` => `CountingBase`
- Add pullbacks
- Add mass interface (fixes https://github.com/cscherrer/MeasureBase.jl/issues/82, fixes https://github.com/cscherrer/MeasureTheory.jl/issues/128)
- Fixes https://github.com/cscherrer/MeasureBase.jl/issues/92
- Fixes https://github.com/cscherrer/DistributionMeasures.jl/issues/
- Improve type stability for transports (https://github.com/cscherrer/MeasureBase.jl/pull/88/commits/e9cffcdd02608edf705af84ee8faf6fd6ebd6865 - thanks @oschulz )
- Make `Density` and `LogDensity` callable (https://github.com/cscherrer/MeasureBase.jl/pull/85)
- Add `smf` and `invsmf` (https://github.com/cscherrer/MeasureBase.jl/pull/100)


cscherrer:
There's currently an error in tests. This correctly throws an error:
```julia
julia> transport_to(StdNormal(), StdUniform())([0.3, 0.4])
ERROR: ArgumentError: Invalid variate type for measure
Stacktrace:
 [1] checked_arg
   @ ~/git/MeasureBase.jl/src/primitives/lebesgue.jl:49 [inlined]
 [2] _default_checked_arg
   @ ~/git/MeasureBase.jl/src/getdof.jl:75 [inlined]
 [3] checked_arg
   @ ~/git/MeasureBase.jl/src/getdof.jl:79 [inlined]
 [4] (::MeasureBase.TransportFunction{StdNormal, StdUniform})(x::Vector{Float64})
   @ MeasureBase ~/git/MeasureBase.jl/src/transport.jl:211
 [5] top-level scope
   @ REPL[71]:1
```

But `StdNormal` does not yet behave properly:
```julia
julia> transport_to(StdNormal(), StdNormal())([0.3, 0.4])
MeasureBase.NoArgCheck{MeasureBase.LebesgueMeasure, Vector{Float64}}
```

oschulz:
I think we need

```julia
@inline MeasureBase.basemeasure(::StdNormal) = MeasureBase.WeightedMeasure(static(-0.5 * log2π), MeasureBase.Lebesgue())
```

So `Lebesgue()` instead of `LebesgueMeasure()`, otherwise `checked_arg` doesn't work. We also have `Lebesgue()` as the base measure for the other std measures.

oschulz:
Question, do we really need the intermediate `WeightedMeasure(static(-0.5 * log2π), ...)`? This might actually be faster, doing everything in one go

```
@inline logdensity_def(::StdNormal, x::U) where {U<:Real} = muladd(abs2(x), -U(1)/U(2), -log2π/U(2))
@inline basemeasure(::StdNormal) = Lebesgue()
```

since it can use hardware-based fused-multiply-add.

cscherrer:
> So `Lebesgue()` instead of `LebesgueMeasure()`, otherwise `checked_arg` doesn't work. We also have `Lebesgue()` as the base measure for the other std measures.

Previously, a lot of measures had `Lebesgue()` as their base measure, but `basemeasure(Lebesgue()) == LebesgueMeasure()`. This seems bad, because `Lebesgue()` and `LebesgueMeasure()` are functionally identical. The extra step was not immediately a problem, but it does make the `basemeasure_sequence` one step longer, for no real benefit. So if we build measures on measures on measures, my guess is that it would lead to compiler issues a little earlier.

The idea with separating `::Lebesgue` from `::LebesgueMeasure` in the first place was that the former can take a "domain" argument. Then `inbounds(Lebesgue(dom))` acts like `RestrictedMeasure(x -> in(x, dom), LebesgueMeasure())`. 

I think there are two possible solutions:
1. Use `LebesgueMeasure()` as the base measure wherever possible, and leave `Lebesgue` as a more flexible user-facing alternative. 
2. Drop `LebesgueMeasure` altogether, and use `Lebesgue{()}` in its place. It's still determined entirely by its type. 

> Question, do we really need the intermediate `WeightedMeasure(static(-0.5 * log2π), ...)`?

It would be convenient to change `MeasureTheory.Normal` to build on top of `StdNormal` instead of starting from scratch. The point of the `WeightedMeasure` is that if you have, say, a regression problem with `Normal` noise, the product of `Normal` measures will have a common base measure, which allows you to only compute it once. In MeasureTheory, this approach gets us just a little faster than Distributions, even for a single evaluation. I think the advantage might come from the `static`ness, but I'm not sure.

Anyway, we could sacrifice this convenience if `fma` gets us faster still. Maybe we should have a small collection of benchmarks to help guide us on things like this?

cscherrer:
Also, ...

> So `Lebesgue()` instead of `LebesgueMeasure()`, otherwise `checked_arg` doesn't work. We also have `Lebesgue()` as the base measure for the other std measures.

I added `@inline checked_arg(::LebesgueMeasure, x::Real) = x`. That got rid of lots of errors, but there's still the one I described above, which I don't yet understand.

cscherrer:
Right, this one has been kind of tricky. I think the invariant we need is that is `mu << LebesgueMeasure()` and `nu << LebesgueMeasure()`, then
```julia
logdensity_rel(mu, nu) == logdensity_rel(mu, LebesgueMeasure()) - logdensity_real(nu, LebesgueMeasure())
```

Say `mu` is a measure on the `x` axis, and `nu` is a measure on the `y` axis. Then the result we want is that they should be singular at the origin. Having LebesgueMeasure dominate both will cause problems. This makes me think maybe we should drop the idea that LebesgueMeasure is arbitrary, and only use it for measures on the real line.

@mschauer , do you have any thoughts on this?

cscherrer:
@oschulz I don't understand how `checked_arg` works, could you point me in the right direction? I'm getting
```julia
julia> using MeasureBase

julia> using Test

julia> using MeasureBase: getdof, check_dof, checked_arg

julia> μ0 = StdExponential()
StdExponential()

julia> x0 = rand(μ0)
0.8225907598317117

julia> μ2 = StdExponential()^(2, 3)
StdExponential() ^ (2, 3)

julia> x2 = rand(μ2)
2×3 Matrix{Float64}:
 1.59882   3.697    0.986999
 0.835304  0.80118  1.03364

julia>     @test_throws ArgumentError checked_arg(μ0, x2)
Test Failed at REPL[25]:1
  Expression: checked_arg(μ0, x2)
    Expected: ArgumentError
  No exception thrown
ERROR: There was an error during testing

julia> checked_arg(μ0, x2)
MeasureBase.NoArgCheck{MeasureBase.LebesgueMeasure, Matrix{Float64}}
```

src/combinators/transformedmeasure.jl
```diff
@@ -86,9 +103,14 @@ end
...
-@inline from_origin(ν::PushforwardMeasure, x) = ν.f(x)
-@inline to_origin(ν::PushforwardMeasure, y) = ν.inv_f(y)
+# @inline transport_origin(ν::PushforwardMeasure) = ν.origin
+# @inline from_origin(ν::PushforwardMeasure, x) = ν.f(x)
+# @inline to_origin(ν::PushforwardMeasure, y) = ν.inv_f(y)
+
+@inline transport_origin(μ::PushforwardMeasure) = transport_origin(parent(μ))
+@inline from_origin(μ::PushforwardMeasure, x) = μ.f(from_origin(parent(μ), x))
+@inline to_origin(μ::PushforwardMeasure, y) = μ.inv_f(to_origin(parent(μ), y))
+
```
cscherrer, line 113: @oschulz what do you think of transporting pushforwards to the origin of the parent like this, instead of stopping at the parent?

oschulz: That kind of forwarding should happen automatically, by the transport "resolver". I have some improvement on that in the works on the "typestable-transport" branch (will be back on this now).

oschulz:
> This makes me think maybe we should drop the idea that LebesgueMeasure is arbitrary, and only use it for measures on the real line.

To only one Lebesgue measure type? Yes, I think that would be less confusing. If we want only half of the real line or so, we can have a "restricted measure" wrapper or so.

mschauer:
I think the type system might not be expressive enough to have always the domain encoded into the Lebesgue measure: you'll have to do type arithmetic like the marginal of a SVector{2} lebesgue measure is a SVector{1} lebegue measure and that's something which in Julia at some point gets in the way of productivity.

src/combinators/transformedmeasure.jl
cscherrer: I see, so it sounds like you're recommending in general we only take a small step and let the resolver figure it out. Is that right?

cscherrer:
In the current setup:
1. We have a singleton type for the most common "Lebesgue measure on $\mathbb{R}$". This is currently called `LebesgueMeasure`
2. We have a convenience wrapper so users can construct "Lebesgue measure on ____". This is currently called `Lebesgue`. In some cases this acts like a `restrict` call, as in `Lebesgue(𝕀)`. In others in can be something completely different, like `Lebesgue(Simplex())`

We need (1) as a common base measure for any restricted Lebesgue measure, or as a root measure for any univariate continuous measure.

(2) is a convenience, but without it users would have to write things like `restrict(x -> 0<x<1, LebesgueMeasure)`. This gets even messier for things like "Lebesgue measure on a simplex".

I agree the naming is weird. Maybe we should change `LebesgueMeasure` to `LebesgueBase` to hint its intended use? Or we could talk about more fundamental refactoring if you have ideas.

src/combinators/transformedmeasure.jl
```diff
@@ -91,15 +108,38 @@ end
...
+measure](https://en.wikipedia.org/wiki/Pushforward_measure) from `μ` the
+[measurable function](https://en.wikipedia.org/wiki/Measurable_function) `f`.
+
+If `f_inverse` is specified, it must be a valid inverse of the function given by
+restricting `f` to the support of `μ`.
 """
-pushfwd(f, μ, volcorr = WithVolCorr()) = PushforwardMeasure(f, inverse(f), μ, volcorr)
+pushfwd(f, μ, volcorr::TransformVolCorr = WithVolCorr()) = PushforwardMeasure(f, inverse(f), μ, volcorr)
+
+pushfwd(f, finv, μ, volcorr::TransformVolCorr = WithVolCorr()) = PushforwardMeasure(f, finv, μ, volcorr)
```
cscherrer, line 128: Not working yet:
```julia
julia> μ = Lebesgue() ^ 2
Lebesgue(ℝ) ^ 2

julia> f(x) = [2x[1], x[2]]
f (generic function with 1 method)

julia> finv(y) = [y[1]/2, y[2]]
finv (generic function with 1 method)

julia> logdensityof(pushfwd(f, finv, μ), zeros(2))
ERROR: MethodError: no method matching iterate(::ChangesOfVariables.NoLogAbsDetJacobian{typeof(finv), Vector{Float64}})
Closest candidates are:
  iterate(::Union{LinRange, StepRangeLen}) at range.jl:872
  iterate(::Union{LinRange, StepRangeLen}, ::Integer) at range.jl:872
  iterate(::T) where T<:Union{Base.KeySet{<:Any, <:Dict}, Base.ValueIterator{<:Dict}} at dict.jl:712
  ...
Stacktrace:
 [1] indexed_iterate(I::ChangesOfVariables.NoLogAbsDetJacobian{typeof(finv), Vector{Float64}}, i::Int64)
   @ Base ./tuple.jl:91
 [2] logdensity_def
   @ ~/git/MeasureBase.jl/src/combinators/transformedmeasure.jl:62 [inlined]
 [3] unsafe_logdensityof
   @ ~/git/MeasureBase.jl/src/density.jl:139 [inlined]
 [4] logdensityof(μ::PushforwardMeasure{typeof(f), typeof(finv), PowerMeasure{Lebesgue{MeasureBase.RealNumbers}, Tuple{Base.OneTo{Int64}}}, MeasureBase.WithVolCorr}, x::Vector{Float64})
   @ MeasureBase ~/git/MeasureBase.jl/src/density.jl:123
 [5] top-level scope
   @ REPL[46]:1
```

src/combinators/transformedmeasure.jl
oschulz: Yes, exactly.

oschulz: And it has a barrier against infinite recursion built in.

src/combinators/transformedmeasure.jl
cscherrer: It works!!
```julia
julia> using MeasureBase, ForwardDiff, ChangesOfVariables, IrrationalConstants

julia> μ = Lebesgue(MeasureBase.BoundedReals(-halfπ, halfπ))
Lebesgue(MeasureBase.BoundedReals{Float64, Irrational{:halfπ}}(-1.5707963267948966, halfπ))

julia> function f(x)
           asin(x)
       end
f (generic function with 1 method)

julia> function finv(y)
           if -halfπ ≤ y ≤ halfπ
               return sin(y)
           else
               @error "finv is only defined on [-π/2, π/2]"
           end
       end
finv (generic function with 1 method)

julia> function withlogjac(f, x)
           dx = ForwardDiff.Dual{ForwardDiff.Tag{typeof(f)}}(x, 1.0)
           dy = f(dx)
           value = dy.value
           deriv = first(dy.partials)
           (value, log(abs(deriv)))   
       end
withlogjac (generic function with 1 method)

julia> function ChangesOfVariables.with_logabsdet_jacobian(::typeof(f), x)
           withlogjac(f, x)
       end

julia> function ChangesOfVariables.with_logabsdet_jacobian(::typeof(finv), x)
           withlogjac(finv, x)
       end

julia> # Both of these pass
       ChangesOfVariables.test_with_logabsdet_jacobian(f, 2 * rand() - 1, ForwardDiff.derivative)
Test Summary:                                                 | Pass  Total  Time
test_with_logabsdet_jacobian: f with input 0.9708087843236501 |    2      2  0.0s

julia> ChangesOfVariables.test_with_logabsdet_jacobian(finv, π * rand() - halfπ, ForwardDiff.derivative)
Test Summary:                                                     | Pass  Total  Time
test_with_logabsdet_jacobian: finv with input 0.43781065021456067 |    2      2  0.0s

julia> ν = pushfwd(f, finv, μ)
PushforwardMeasure(
    f,
    finv,
    Lebesgue(MeasureBase.BoundedReals{Float64, Irrational{:halfπ}}(-1.5707963267948966, halfπ)))

julia> logdensityof(ν, π/4)
-0.3465735902799726
```

cscherrer:
I got the `pushfwd(f, finv, μ)` for working, but I noticed some issues along the way:
https://github.com/cscherrer/MeasureBase.jl/issues/89

src/combinators/transformedmeasure.jl
mschauer: 🎉 juhu!

oschulz:
Can we *not* add `pushfwd(f, finv, μ)`? I think we can make JuliaMath/InverseFunctions.jl#18 happen.

cscherrer:
I agree it's better to solve the problem once in InverseFunctions than for every user to replicate the work. But for users of MeasureBase, `pushfwd(f, finv, μ)` still seems much better than `pushfwd(setinverse(f, finv), μ)`. We could have a single line to make this definition, and save everyone downstream some typing. Do you see problem with this approach?

oschulz:
> We could have a single line to make this definition, and save everyone downstream some typing.We could have a single line to make this definition, and save everyone downstream some typing.

Oh, you mean this would do

```
pushfwd(f, finv, μ) = pushfwd(setinverse(f, finv), μ)
```

Once we have that capability in InverseFunctions? Yes, that would be very user friendly, I like it.

oschulz:
Regarding `massof` - I think it would be more natural if it returns a `Real` (or static real) if the mass is known, and `UnknownFiniteMass` resp. `UnknownMass` otherwise.

oschulz:
`isnormalized` I would see as something that only operates on measures, not on return values of `massof`. It's also a tricky choice of function name, since we probably shouldn't export something that generic.

oschulz:
Regarding `normalize(m::AbstractMeasure) = normalize(m, massof(m))` - this may not be ideal, since the second argument of `LinearAlgebra.normalize` has very different semantics by default (indicating the type of norm, not the current norm value). Maybe that should be `normalize(m::AbstractMeasure) = renormalize(m, 1/massof(m))` or no?

cscherrer:
> Regarding `massof` - I think it would be more natural if it returns a `Real` (or static real) if the mass is known, and `UnknownFiniteMass` resp. `UnknownMass` otherwise.

We could take this a step further and use `missing` instead of `UnknownMass()`. But then it seems strange to only have `UnknownFiniteMass` as special.

> `isnormalized` I would see as something that only operates on measures, not on return values of `massof`.

I agree. What I have is just for dispatch, but there it should probably be `_isnormalized`.

> It's also a tricky choice of function name, since we probably shouldn't export something that generic.

I think my position on this is the same as my position on software patents. It would be better if no one did it. But they do, and if you don't you're only hurting yourself. 

MeasureBase is a pretty fundamental package. If someone else wants a name for a non-measure use we can talk about splitting out an interface package.

> Regarding `normalize(m::AbstractMeasure) = normalize(m, massof(m))` - this may not be ideal, since the second argument of `LinearAlgebra.normalize` has very different semantics by default (indicating the type of norm, not the current norm value). Maybe that should be `normalize(m::AbstractMeasure) = renormalize(m, 1/massof(m))` or no?

This is a good point, especially if we make `massof` sometimes return a float.

oschulz:
> We could take this a step further and use missing instead of UnknownMass(). But then it seems strange to only have UnknownFiniteMass as special.

Hm, yes. One advantage of `UnknownMass{MeasureType}()` would be that it helps more with debugging than missing, e.g. when code tries to do math with the result of `massof`, you'd get an error that includes the measure type (even if the value has been passed around already).

oschulz:
>> isnormalized ... we probably shouldn't export something that generic
> MeasureBase is a pretty fundamental package

True, but `normalize` is defined in `LinearAlgebra`, which is another level of fundamental. So if we export `isnormalized`, it would fall to us to also support it for arrays and so on, since no other package could do so without committing type piracy ...

cscherrer:
> True, but `normalize` is defined in `LinearAlgebra`, which is another level of fundamental. So if we export `isnormalized`, it would fall to us to also support it for arrays and so on, since no other package could do so without committing type piracy ...

I see, people would expect other methods just because of the name. Well, I think it's almost trivial to provide that in an extensible way. We just set
```julia
isnormalized(x, p::Real=2) = isone(norm(x, p))
```

Or we could just not export it, if you think it's a bad idea.

src/mass-interface.jl
```diff
@@ -0,0 +1,75 @@
...
+        @eval begin
+            Base.$op(::$U, ::$UF) = $U()
+            Base.$op(::$UF, ::$U) = $U()
+        end
+    end
+end
+
+massof(m::AbstractMeasure) = UnknownMass(m)
+
+logmassof(m::AbstractMeasure) = log(massof(m))
```
cscherrer, line 37:
```suggestion
```

Maybe we should wait on this? It will require a separate version of the above types and methods, which seems kind of heavy for relatively little benefit. Not sure yet about this.

cscherrer:
You can now do e.g.
```julia
julia> (MeasureBase.StdNormal())(2..4)
0.02271846070634609
```
@oschulz pointed out that while this works well for `<:AbstractMeasure`s, this is difficult to generalize. For example, DistributionMeasures helps `Distribution`s satisfy the measure interface, but we can't just make them callable. We'll need another option. For example, we could introduce a two-argument `massof` method.

src/combinators/powerweighted.jl
```diff
@@ -35,3 +35,5 @@ end
 function Pretty.tile(d::PowerWeightedMeasure)
     Pretty.pair_layout(Pretty.tile(d.parent), Pretty.tile(d.exponent), sep = " ↑ ")
 end
+
+massof(m::PowerWeightedMeasure) = massof(m.parent)^m.exponent
```
cscherrer, line 39: This is wrong

cscherrer:
I think this might be pretty close. We still need to update the transport definitions for weighted measures. I think it's important for mass to be invariant under transports; otherwise it's easy for things to get mixed up. With invariant mass we'll be able to use the transport API to get the measure of an interval, basically an extension of the usual `cdf` on unit-mass measures.

oschulz:
> I think it's important for mass to be invariant under transports

I agree.

cscherrer:
I think I see the problem. I changed it back here:
https://github.com/cscherrer/MeasureBase.jl/pull/88/commits/d0530c3bc2891243d21a052139aaeb9d99f381a8

But now there's nothing enforcing that transports should preserve mass, so we'll need to figure that out. @oschulz I think you had a suggestion for this?

src/combinators/transformedmeasure.jl
```diff
@@ -88,18 +115,76 @@ end
...
+
+function _pushfwd(f::FunctionWithInverse, μ, ::V, v::V) where {V}
+    pushfwd(setinverse(f.f ∘ μ.f, μ.finv ∘ f.invf), μ.origin, v)
+end
+
+function _pushfwd(f, μ, _, v)
+    pushfwd(setinverse(f, inverse(f)), μ, v)
+end
+
+getdof(μ::PushforwardMeasure) = getdof(transport_origin(μ))
```
cscherrer, line 168:
```suggestion
```
Duplicate method

src/combinators/transformedmeasure.jl
```diff
@@ -17,32 +19,58 @@ function parent(::AbstractTransformedMeasure) end
...
+
+    Users should not call `PushforwardMeasure` directly. Instead call or add
+    methods to `pushfwd`.
 """
-struct PushforwardMeasure{FF,IF,M,VC<:TransformVolCorr} <: AbstractPushforward
-    f::FF
-    inv_f::IF
+struct PushforwardMeasure{F,I,M,VC<:TransformVolCorr} <: AbstractPushforward
+    f::F
+    finv::I
```
oschulz, line 30: Why not keep `inv_f`, as before?

src/combinators/transformedmeasure.jl
```diff
@@ -17,32 +19,58 @@ function parent(::AbstractTransformedMeasure) end
...
     origin::M
     volcorr::VC
 end
 
 gettransform(ν::PushforwardMeasure) = ν.f
 parent(ν::PushforwardMeasure) = ν.origin
 
+function transport_def(ν::PushforwardMeasure{F,I,M}, μ::M, x) where {F,I,M}
```
oschulz, line 38: Do we need `transport_def` for `PushforwardMeasure`? It shouldn't be necessary since `PushforwardMeasure` should use the "origin path" for transport.

src/combinators/transformedmeasure.jl
```diff
@@ -57,49 +85,99 @@ end
...
-from `μ` the [measurable function](https://en.wikipedia.org/wiki/Measurable_function) `f`.
+Return the [pushforward
+measure](https://en.wikipedia.org/wiki/Pushforward_measure) from `μ` the
+[measurable function](https://en.wikipedia.org/wiki/Measurable_function) `f`.
+
+To manually specify an inverse, call 
+`pushfwd(InverseFunctions.setinverse(f, finv), μ, volcorr)`.
+"""
+function pushfwd(f, μ, volcorr::TransformVolCorr = WithVolCorr())
+    pushfwd(setinverse(f, inverse(f)), μ, volcorr)
```
oschulz, line 139: This looks weird - if `f` has an inverse we shouldn't need to set it's inverse.

oschulz (changes requested):

src/combinators/transformedmeasure.jl
cscherrer: It's for uniformity. The main constructor is
```julia
function pushfwd(f::FunctionWithInverse, μ, volcorr::TransformVolCorr)
    PushforwardMeasure(f, inverse(f), μ, volcorr)
end
```

Other calls get passed to this, so we can assume `f` is always a `FunctionWithInverse`. 

Mostly it's nice to avoid having to worry about lots of special cases. In the last implementation of this, I had that `f` and `finv` were _never_ `FunctionWithInverse` - if it saw one of those it would pull it apart. If we don't want that, the best alternative I see is to _always_ have `FunctionWithInverse`. 

The only other possibility is to just let everything be arbitrary. In that case construction is easy, but any use of a `PushforwardMeasure` would need to have methods for the different cases. This seems very awkward to me. If you think it could be better, maybe a PR into `dev` and we can discuss?

src/combinators/transformedmeasure.jl
cscherrer: That's very awkward to me. We talk about "inverse CDF" or "inverse sine / arcsin". But for functions with a generic name, we usually say "f" and "f inverse".

src/combinators/transformedmeasure.jl
cscherrer: Good catch! I'll update this, let's wait to resolve this comment until I do.

src/combinators/transformedmeasure.jl
oschulz: Sure, I don't have a strong opinion here.

src/standard/stdlogistic.jl
```diff
@@ -5,11 +5,17 @@ export StdLogistic
 @inline insupport(d::StdLogistic, x) = true
 
 @inline logdensity_def(::StdLogistic, x) = (u = -abs(x); u - 2 * log1pexp(u))
-@inline basemeasure(::StdLogistic) = Lebesgue()
+@inline basemeasure(::StdLogistic) = LebesgueBase()
 
-@inline transport_def(::StdUniform, μ::StdLogistic, x) = logistic(x)
-@inline transport_def(::StdLogistic, μ::StdUniform, x) = logit(x)
+@inline transport_def(::StdUniform, μ::StdLogistic, x) = smf(StdLogistic(), x)
```
oschulz, line 10: This indirection may cause a longer tape with Zygote

src/combinators/transformedmeasure.jl
```diff
@@ -57,49 +69,93 @@ end
...
-pushfwd(f, μ, volcorr = WithVolCorr()) = PushforwardMeasure(f, inverse(f), μ, volcorr)
+function pushfwd(f, μ, volcorr::TransformVolCorr = WithVolCorr())
+    PushforwardMeasure(f, inverse(f), μ, volcorr)
+end
+
+function pushfwd(f, μ::PushforwardMeasure, volcorr::TransformVolCorr = WithVolCorr())
+    _pushfwd_of_pushfwd(f, μ, μ.volcorr, volcorr)
+end
+
+function pushfwd(f, μ::PushforwardMeasure, ::WithVolCorr)
```
oschulz, line 128: Is this necessary to resolve dispatch ambiguities?

src/combinators/transformedmeasure.jl
```diff
@@ -57,49 +69,93 @@ end
...
+    _pushfwd_of_pushfwd(f, μ, μ.volcorr, WithVolCorr())
+end
+
+# Either both WithVolCorr or both NoVolCorr, so we can merge them
+function _pushfwd_of_pushfwd(f, μ::PushforwardMeasure, ::V, v::V) where {V}
+    pushfwd(f ∘ μ.f, μ.origin, v)
```
oschulz, line 134: Maybe use `FunctionChains.fchain((μ.f, f))` here until Julia v1.9 is released, since chained `∘` is not type-stable in v1.8 (but is type stable in v1.6).

PR #94, opened 2022-09-30
Transport-exploration

cscherrer:
This PR explores the possibility of implementing the transport API in a parallel way to the way we handle calls to `logdensity_rel`. Some history is probably helpful here...

The current transport implementation is the work of @oschulz . In trying to understand and extend this, I ran into a few things that were confusing to me. 

First, the concept of a "transport origin". I had assumed this would be analogous to `rootmeasure`. I think this is just my interpretation of the work "origin". To me that word has a sense of being absolute, so "origin of an origin" doesn't really make sense. In terms of an implementation, this isn't such a big problem - we can either clarify the term or change it without affecting the structure of the code.

This first difficulty also bit me when trying to adjust the behavior of transports for weighted measures. I had expected we could just do
```julia
transport_origin(ν::WeightedMeasure) = weightedmeasure(ν.logweight, transport_origin(ν.base))
to_origin(w::WeightedMeasure, y) = to_origin(w.base, y)
from_origin(w::WeightedMeasure, x) = from_origin(w.base, x)
```

But this breaks, because it tries to make calls like `from_origin(::StdNormal, x)`. I found this very confusing.

Second, trying to add transport methods kept leading me into errors like "no method matching `foo(::NoTransportOrigin)`". It was very challenging to unravel where this error was coming from. This can probably be improved by intercepting such calls earlier and adding a helpful error message.

Finally, @oschulz had mentioned that there were some lingering issues of type instability with the current approach.

It struck me that abstractly, the `transport_to` problem is exactly the same as the `logdensity_rel` problem. Here's the correspondence:

interfaces

Relative log-densities were type-unstable for a very long time. If we finally have a solution (I think we do) and we now come across a new problem of type instability, it seems sensible to take the same approach. In terms of MeasureBase in general, this is also appealing, since new devs would only have to learn the dispatch patterns once. Having understood `logdensity_rel`, transports would quickly come to feel familiar.

One obvious problem with this is that @oschulz put a lot of energy into the current approach before I realized any of this. So let's take https://github.com/cscherrer/MeasureBase.jl/pull/94 more as a point of discussion than something I expect to see merged soon. Maybe it will be a helpful solution, or maybe the current approach with some added docs and error messages will serve us better.


src/standard/stdexponential.jl
```diff
@@ -5,9 +5,11 @@ export StdExponential
...
 @inline logdensity_def(::StdExponential, x) = -x
-@inline basemeasure(::StdExponential) = Lebesgue()
-
-@inline transport_def(::StdUniform, μ::StdExponential, x) = -expm1(-x)
-@inline transport_def(::StdExponential, μ::StdUniform, x) = -log1p(-x)
+@inline basemeasure(::StdExponential) = LebesgueBase()
 
 Base.rand(rng::Random.AbstractRNG, ::Type{T}, ::StdExponential) where {T} = randexp(rng, T)
+
+transport_origin(::StdExponential) = StdUniform()
```
oschulz, line 12: Standard measures shouldn't have an origin. The "transport resolver" uses origins first and then falls back to `transport_def`, we don't want everything to be forced through uniform.

src/standard/stdnormal.jl
```diff
@@ -0,0 +1,21 @@
...
+
+@inline logdensity_def(::StdNormal, x) = -x^2 / 2
+@inline basemeasure(::StdNormal) = WeightedMeasure(static(-0.5 * log2π), LebesgueBase())
+
+@inline getdof(::StdNormal) = static(1)
+
+
+@inline Base.rand(rng::Random.AbstractRNG, ::Type{T}, ::StdNormal) where {T} = randn(rng, T)
+
+transport_origin(::StdNormal) = StdUniform()
```
oschulz, line 18: Standard measures shouldn't have an origin, see above.

src/transport.jl
```diff
@@ -16,7 +6,7 @@ between `ν` and another measure.
 """
 function transport_origin end
 
-transport_origin(ν::NU) where {NU} = NoTransformOrigin{NU}()
+transport_origin(ν::NU) where {NU} = ν
```
oschulz, line 9: I really don't like measures being their own origin, it break the semantics of what origin means. And it's much more clean to check for `NoTransportOrigin` than to compare a measure with itself (also potentially faster and more type stable if the measure is parameterized).

cscherrer: This meaning of "origin" is very counter-intuitive to me, and with the current setup `NoTransportOrigin` propagates through the code. The problem with this is that error messages can be very strange, like "no `foo` method for `NoTransportOrigin`".

If this otherwise works well, this could be ok as long as we have docs to describe the behavior, and we can improve the error messages when people extend the system and don't quite get it perfect on the first attempt.

oschulz: > This meaning of "origin" is very counter-intuitive to me

We can rename it - the semantics is the one we need though: Is there a preferred/natural measure that this measure should transform from/to?

`Normal`, for example, shouldn't have an origin, instead it should have `transport_def`-implementation to transform to/from both standard normal and uniform - so that in a "go to gaussian space" use case it won't be forced through a uniform with larger numerical errors at the extreme values. A pushforward measure, on the other hand, definitely has exactly one measure that values should be transported to first when resolving a transport.

> The problem with this is that error messages can be very strange

Depends - it still gives information about the measure that didn't have an origin. An users typically won't see `NoTransportOrigin` if they don't use the origin API directly, they should get a `NoTransport` instead when they use `transport_to`.

PR #95, opened 2022-09-30
Improve transport implementation and add product support

oschulz:
Increases type stability.

### DIFF
--- a/src/transport.jl
+++ b/src/transport.jl
@@ -113,9 +113,6 @@ See [`transport_to`](@ref).
 """
 function transport_def end
 
-transport_def(::Any, ::Any, x::NoTransportOrigin) = x
-transport_def(::Any, ::Any, x::NoTransport) = x
-
 function transport_def(ν, μ, x)
     _transport_between_origins(ν, _origin_depth(ν), _origin_depth(μ), μ, x)
 end
@@ -175,9 +172,13 @@ end
 @inline _transport_intermediate(::Integer, n_μ::Integer) = StdUniform()^n_μ
 @inline _transport_intermediate(::StaticInt{1}, ::StaticInt{1}) = StdUniform()
 
+_call_transport_def(ν, μ, x) = transport_def(ν, μ, x)
+_call_transport_def(::Any, ::Any, x::NoTransportOrigin) = x
+_call_transport_def(::Any, ::Any, x::NoTransport) = x
+
 function _transport_with_intermediate(ν, m, μ, x)
-    z = transport_def(m, μ, x)
-    y = transport_def(ν, m, z)
+    z = _call_transport_def(m, μ, x)
+    y = _call_transport_def(ν, m, z)
     return y
 end
 
@@ -218,7 +219,7 @@ function Base.:(==)(a::TransportFunction, b::TransportFunction)
 end
 
 Base.@propagate_inbounds function (f::TransportFunction)(x)
-    return transport_def(f.ν, f.μ, checked_arg(f.μ, x))
+    return _call_transport_def(f.ν, f.μ, checked_arg(f.μ, x))
 end
 
 @inline function InverseFunctions.inverse(f::TransportFunction{NU,MU}) where {NU,MU}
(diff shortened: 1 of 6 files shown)


oschulz:
@cscherrer , can we get this one one first?

src/transport.jl
```diff
@@ -113,54 +113,64 @@ See [`transport_to`](@ref).
...
 
 function transport_def(ν, μ, x)
-    _transport_with_intermediate(
-        ν,
-        _checked_transport_origin(ν),
-        _checked_transport_origin(μ),
-        μ,
-        x,
-    )
+    _transport_between_origins(ν, _origin_depth(ν), _origin_depth(μ), μ, x)
```
cscherrer, line 120: This seems to be greedy about origins. Often there might be no `transport_def` method between two measures, but there _can_ be one between origins of those measures. Ideally, we want to use `to_origin` and `from_origin` as few times as possible.

In `logdensity_rel`, we do this in terms of `commonbase`. Given the `basemeasure_sequence` of each measure, there may be lots of pairs of depths allowing us to "jump across" from one sequence to the other. This method helps us find the optimal depths:
```julia
@generated function commonbase(μ::M, ν::N, ::Type{T}) where {M<:Tuple,N<:Tuple,T}
    m = schema(M)
    n = schema(N)

    sols = Iterators.filter(
        ((i, j),) -> static_hasmethod(logdensity_def, Tuple{m[i],n[j],T}),
        Iterators.product(1:length(m), 1:length(n)),
    )
    isempty(sols) && return :(nothing)
    minsol = static.(argmin(((i, j),) -> i + j, sols))
    quote
        $minsol
    end
end
```

oschulz:
Ok, we have initial support for products now with this:

```julia
μ = productmeasure((a = StdExponential(), b = StdLogistic()^3))
ν = StdUniform()^(2, 2)
x = rand(μ)
y = transport_to(ν, μ)(x)
transport_to(μ, ν)(y)
```

src/transport.jl
oschulz: > This seems to be greedy about origins.

It is. But only measure that have a clear origin should define `transport_origin`, so I think it's Ok.

oschulz: The situation is a bit different from `logdensity_rel` because in basically all practical use cases one side of the transport will  be a power of a standard dist.

cscherrer: Great point. It's possible people might want to use it in other cases, but it makes sense to keep it more direct like this until those cases come up in practice, if they ever do.

oschulz:
Ok, some type inference failures on nightly, but I think we can ignore those for now, that's not uncommon.

cscherrer:
Great! We should also add 1.8 to the CI

src/transport.jl
```diff
@@ -34,7 +34,7 @@ Pull `y` from `ν` back to `MeasureBase.transport_origin(ν)`.
 """
 function to_origin end
 
-to_origin(ν::NU, ::Any) where {NU} = NoTransformOrigin{NU}(ν)
+to_origin(ν::NU, ::Any) where {NU} = NoTransportOrigin{NU}(ν)
```
cscherrer, line 37: Typo? `NoTransportOrigin` doesn't take args
```suggestion
to_origin(ν::NU, ::Any) where {NU} = NoTransportOrigin{NU}()
```

oschulz:
> What do you think of merging this into dev and getting transports working right on WeightedMeasures before merging into master

Good idea, go right ahead.

src/transport.jl
oschulz: Ooops, yes, typo - thanks!

cscherrer:
@oschulz when you get a chance could you have a look at the merge conflicts for merging this into `dev`? Your code, so I think it's better if you make the call how to resolve

cscherrer (approved):

oschulz:
> have a look at the merge conflicts for merging this into dev? 

Done (I hope correctly): #97 replaces this PR.

PR #96, opened 2022-10-01
density-core

cscherrer:
I'd like to get all the new stuff into https://github.com/cscherrer/MeasureBase.jl/pull/88 before merging into master. But it looks like https://github.com/cscherrer/MeasureBase.jl/pull/85 makes this a little tricky. In addition to code changes, it also reorganizes `density.jl`, breaking the "core" stuff into `density-core.jl`. That's good, but very awkward to do _after_ other changes.

This little PR instead makes _only_ the reorganization changes - the functionality is exactly the same as master.

I think we should merge this into master first. Then getting other code in line should be easier.


src/density.jl
```diff
@@ -100,173 +100,9 @@ Define a new measure in terms of a log-density `f` over some measure `base`.
...
-export logdensityof
-
-export density_def
-
-density_def(μ, ν::AbstractMeasure, x) = exp(logdensity_def(μ, ν, x))
-density_def(μ, x) = exp(logdensity_def(μ, x))
 
 """
-    rebase(μ, ν)
+rebase(μ, ν)
```
cscherrer, line 105:
```suggestion
    rebase(μ, ν)
```

oschulz (approved):
This basically reorganises the `logdensityof` implementation, right? I'm happy with it if you are. :-)

PR #98, opened 2022-10-03
Pushfwd-inverses

cscherrer:
Requires `InverseFunctions = "0.1.8"`, and reworks `pushfwd(f, finv, m)` to use `InverseFunctions.setinverse`


cscherrer:
I think I tracked down the problem:
```julia
julia> μ = StdUniform()
StdUniform()

julia> ν = pushfwd(logit, StdUniform())
PushforwardMeasure(LogExpFunctions.logit, StdUniform())

julia> f = transport_to(ν, μ)
TransportFunction(PushforwardMeasure(LogExpFunctions.logit, StdUniform()), StdUniform())

julia> inverse(f)(y)
logistic (generic function with 2 methods)
```

src/combinators/transformedmeasure.jl
```diff
@@ -17,49 +17,50 @@ function parent(::AbstractTransformedMeasure) end
...
-        invoke(transport_def, Tuple{Any,PushforwardMeasure,Any}, ν, μ, x)
+        invoke(transport_def, Tuple{PushforwardMeasure,Any,Any}, ν, μ, x)
     end
 end
 
-function transport_def(μ::M, ν::PushforwardMeasure{FF,IF,M}, y) where {FF,IF,M}
+function transport_def(μ::M, ν::PushforwardMeasure{F,M}, y) where {F,M}
     if μ == parent(ν)
-        return ν.inv_f(y)
+
```
cscherrer:
```suggestion
```
don't need this extra line

src/combinators/transformedmeasure.jl
```diff
@@ -17,49 +17,46 @@ function parent(::AbstractTransformedMeasure) end
 export PushforwardMeasure
 
 """
-    struct PushforwardMeasure{FF,IF,MU,VC<:TransformVolCorr} <: AbstractPushforward
-        f :: FF
-        inv_f :: IF
+    struct PushforwardMeasure{F,MU,VC<:TransformVolCorr} <: AbstractPushforward
```
oschulz, line 20: Let's keep `inv_f` to cache inverses.

src/combinators/transformedmeasure.jl
```diff
@@ -159,11 +152,11 @@ If `f_inverse` is specified, it must be a valid inverse of the function given by
 restricting `f` to the support of `μ`.
 """
 function pushfwd(f, μ::AbstractMeasure, volcorr::TransformVolCorr = WithVolCorr())
-    pushfwd(f, inverse(f), μ, volcorr)
+    PushforwardMeasure(f, μ, volcorr)
 end
 
 function pushfwd(f, finv, μ::AbstractMeasure, volcorr::TransformVolCorr = WithVolCorr())
```
oschulz, line 158: Let's take `pushfwd(f, finv, μ, ...)` out and only support `pushfwd(f, μ, ...)`. Also, `μ` should not be required to be an `AbstractMeasure`.

cscherrer:
@oschulz we had talked about allowing `f` and `finv` to be `FunctionWithInverse`. But I think it's important to do transformations like unraveling a "pushforward of a pushforward", and this got a little more complex. It was simpler for me to unwrap and re-wrap this more concretely.

oschulz:
Why do we need to unwrap `FunctionWithInverse` - we can just treat it like any other function that has an inverse defined.

cscherrer:
In that case I maybe
```julia
struct PushforwardMeasure{F,I,M,VC<:TransformVolCorr} <: AbstractPushforward
    f::FunctionWithInverse{F,I}
    origin::M
    volcorr::VC
end
```

It would just be nice to have some uniformity in the structure. Then a lot of the code can be simpler

cscherrer:
It's also not clear to me what guarantees InverseFunctions makes. For example, this surprises me
```julia
julia> f = setinverse(sin, asin)
(::InverseFunctions.FunctionWithInverse{typeof(sin), typeof(asin)}) (generic function with 1 method)

julia> finv = setinverse(asin, sin)
(::InverseFunctions.FunctionWithInverse{typeof(asin), typeof(sin)}) (generic function with 1 method)

julia> f ∘ finv
InverseFunctions.FunctionWithInverse{typeof(sin), typeof(asin)}(sin, asin) ∘ InverseFunctions.FunctionWithInverse{typeof(asin), typeof(sin)}(asin, sin)

julia> f ∘ inverse(f)
InverseFunctions.FunctionWithInverse{typeof(sin), typeof(asin)}(sin, asin) ∘ InverseFunctions.FunctionWithInverse{typeof(asin), typeof(sin)}(asin, sin)
```

oschulz:
>  f::FunctionWithInverse{F,I}
> It would just be nice to have some uniformity in the structure.

I think users might prefer have their original function in `mu.f`.

oschulz:
> For example, this surprises me ...

That is the expected behavior.

cscherrer:
> That is the expected behavior.

I would hope there's a possibility of fusing compositions in some cases. We can't do that to Base functions without type piracy. For example, this is wasteful:
```julia
julia> @btime (log ∘ exp)(x) setup=(x=randn())
  11.984 ns (0 allocations: 0 bytes)
0.5002827314874707
```

We ought to be able to do better in cases like this. But maybe that's a discussion for another time

cscherrer:
> I think users might prefer have their original function in `mu.f`. `finv` is really only intended as a caching mechanism, the function supplied by the user should be store in `f`.

Ok, I changed the main constructor to
```julia
function pushfwd(f, μ, volcorr::WithVolCorr)
    PushforwardMeasure(f, inverse(f), μ, volcorr)
end
```

oschulz:
> I would hope there's a possibility of fusing compositions in some cases.

We could, in some cases, but I don't expect that chained `FunctionWithInverse`s will be very common, so I'm not sure how useful it is.

cscherrer:
Found a bug:
```julia
julia> μ = pushfwd(logit, StdUniform())
PushforwardMeasure(LogExpFunctions.logit, StdUniform())

julia> ν = pushfwd(logistic, μ)
PushforwardMeasure(
    InverseFunctions.FunctionWithInverse{ComposedFunction{typeof(logistic), typeof(logit)}, ComposedFunction{typeof(logistic), typeof(logit)}}(LogExpFunctions.logistic ∘ LogExpFunctions.logit, LogExpFunctions.logistic ∘ LogExpFunctions.logit),
    StdUniform())

julia> logdensityof(ν, 0.3)
ERROR: MethodError: no method matching iterate(::ChangesOfVariables.NoLogAbsDetJacobian{InverseFunctions.FunctionWithInverse{ComposedFunction{typeof(logistic), typeof(logit)}, ComposedFunction{typeof(logistic), typeof(logit)}}, Float64})
Closest candidates are:
  iterate(::Union{LinRange, StepRangeLen}) at range.jl:872
  iterate(::Union{LinRange, StepRangeLen}, ::Integer) at range.jl:872
  iterate(::T) where T<:Union{Base.KeySet{<:Any, <:Dict}, Base.ValueIterator{<:Dict}} at dict.jl:712
  ...
Stacktrace:
 [1] indexed_iterate(I::ChangesOfVariables.NoLogAbsDetJacobian{InverseFunctions.FunctionWithInverse{ComposedFunction{typeof(logistic), typeof(logit)}, ComposedFunction{typeof(logistic), typeof(logit)}}, Float64}, i::Int64)
   @ Base ./tuple.jl:91
 [2] logdensityof(ν::PushforwardMeasure{InverseFunctions.FunctionWithInverse{ComposedFunction{typeof(logistic), typeof(logit)}, ComposedFunction{typeof(logistic), typeof(logit)}}, InverseFunctions.FunctionWithInverse{ComposedFunction{typeof(logistic), typeof(logit)}, ComposedFunction{typeof(logistic), typeof(logit)}}, StdUniform, MeasureBase.WithVolCorr}, y::Float64)
   @ MeasureBase ~/git/MeasureBase.jl/src/combinators/transformedmeasure.jl:67
 [3] top-level scope
   @ REPL[18]:1
```

cscherrer:
I got some things straightened out, and I made the tests better. But "Pushforward-of-pushforward" are still all failing. Currently it's
```julia
@testset "transformedmeasure.jl" begin
    for (f, μ, ν_ref) in [
        ((-) ∘ log1p ∘ (-), StdUniform(), StdExponential())
        ((-) ∘ log, StdUniform(), StdExponential())
        (exp ∘ (-), StdExponential(), StdUniform())
        (logit, StdUniform(), StdLogistic())
        (logistic, StdLogistic(), StdUniform())
        (Φ, StdNormal(), StdUniform())
        (Φinv, StdUniform(), StdNormal())
        ]
        
        finv = inverse(f)
        test_pushfwd(f, μ, ν_ref)

        # broken
        @testset "Pushforward-of-pushforward" begin
            # test_pushfwd(finv, pushfwd(f, μ), μ)
        end
    end
end
```

cscherrer:
Getting closer. Current error is
```julia
ArgumentError: Measure ν of type StdUniform has static(1) DOF but μ of type PushforwardMeasure has MeasureBase.NoDOF{PushforwardMeasure{InverseFunctions.FunctionWithInverse{typeof(logit), typeof(LogExpFunctions.logistic)}, InverseFunctions.FunctionWithInverse{typeof(LogExpFunctions.logistic), typeof(logit)}, StdUniform, MeasureBase.WithVolCorr}}() DOF
```

oschulz:
> ERROR: MethodError: no method matching iterate(::ChangesOfVariables.NoLogAbsDetJacobian{InverseFunctions.FunctionWithInverse{ComposedFunction{typeof(logistic), typeof(logit)}, ComposedFunction{typeof(logistic), typeof(logit)}}, Float64})

Ah, darn, we don't have `with_logabsdet_jacobian` support for `FunctionWithInverse` yet, have to add that.

cscherrer:
> Ah, darn, we don't have `with_logabsdet_jacobian` support for `FunctionWithInverse` yet, have to add that.

Yes, but luckily it's pretty easy to work around. Current problem is a DOF mismatch, which seems a little weirder. Looking into that now

cscherrer:
Tests are passing!
```julia
getdof(μ::PushforwardMeasure) = getdof(transport_origin(μ))
```

cscherrer:
Think we need to drop the case
```julia
(f, μ, ν_ref) == ((-) ∘ log1p ∘ (-), StdUniform(), StdExponential())
```
since 1.7 fails type inference for it. Guessing it would do better with FunctionChains.jl

oschulz:
> since 1.7 fails type inference for it. Guessing it would do better with FunctionChains.jl

Should pass again in v1.9 (https://github.com/JuliaLang/julia/issues/45715), but we can definitely use FunctionChains, it shouldn't add any measurable load time.

cscherrer:
I don't think we even need it as a dependency, it was just for one particular test

oschulz:
Ok, then let's maybe use an `if v"1.6" < VERSION < v"1.9"` to run that test without @inferred on v1.8?

cscherrer:
It's passing now, and we'd have to either unwind all of the structure or have a complicated condition to test whether to check `@inferred`. I don't think it would add anything.

The one failing case was on v1.7, and only for `(-) ∘ log1p ∘ (-)`. We can instead check `(-) ∘ log`, and it all works

cscherrer:
Also, a _single_ pushforward using that composed function works just fine. The only problem was in the "pushforward-of-pushforward" test I added, and then only in v1.7

oschulz (approved):

PR #100, opened 2022-10-07
Stieltjes measure function

cscherrer:
The _Stieltjes measure function (SMF)_ is a generalization of the CDF. A function $F$ is the SMF of a measure $\mu$ if 

$$
\mu((a, b]) = F(b) - F(a)
$$

for all $a,b\in \mathbb{R}$.

The inverse of such a function is a function $G$ satisfying

$$
\mu((G(a), G(b)]) = b - a
$$

If $\mu$ is normalized, it can written as $\mu = G_*(u)$, where $u$ is the uniform measure on the unit interval. 

If $\mu$ is dominated by Lebesgue measure, then $F$ and $G$ above are inverses, and $u = F_*(u)$. On the other hand, if $\mu$ is singular, then each point of concentrated measure must be the image under $G$ of some interval. Because of this, $G$ is only a _one-sided_ inverse of $F$: $G \circ F$ is the identity function, but $F \circ G$ is not. 

There are a few things that aren't yet clear to me:
1. The function corresponding to $G$ is currently called `smfinv`. This is a little unsatisfying, since $G$ contains at least as much information as $F$, and sometimes more (since $F$ maps concentrated measure into a point). Is there a better name?
2. There's a natural connection between `smf` and `transport_def(Uniform, mu, x)`. But a method `transport_def(Uniform, mu, x) = smf(mu, x)` doesn't seem quite right. How should we approach this?


cscherrer:
@oschulz I'm not sure yet how to handle the inverse. One idea is to do something like
```julia
smf(mu, x) = smf(mu)(x)
smfinv(mu) = inverse(smf(mu))
smfinv(mu, x) = smfinv(mu)(x)
```

That way could define e.g.
```julia
smf(::StdLogistic) = logistic
```
and get the rest for free.

OTOH, I think I've seen type stability problems in some cases with lots of currying (though I don't have an example on hand). Any thoughts on that?

oschulz:
Hm, might be better to make `smf(mu, x)` and `invsmf(mu,u)` the primary functions for users to specialize and define

```julia
smf(mu) = Base.Fix1(smf, mu)
InverseFunctions.inverse(f::Base.Fix1{typeof(smf)}) = Base.Fix1(invsmf, f.x)

invsmf(mu) = Base.Fix1(invsmf, mu)
InverseFunctions.inverse(f::Base.Fix1{typeof(invsmf)}) = Base.Fix1(smf, f.x)
```

cscherrer:
Ok. That loses us the ability to automate inverses, but that's not too big a deal. I wanted to also be sure AD is still easy - think it would just be
```julia
function smf(μ, x::Dual{TAG}) where TAG
    val = ForwardDiff.value(x)
    Δ = ForwardDiff.partials(x)
    Dual{TAG}(smf(μ, val), Δ * logdensityof(μ, val))
end
```

That `logdensityof` makes me wonder if this is yet another case where things should really be relative. Need to think through that some more

oschulz:
> That densityof makes me wonder if this is yet another case where things should really be relative. Need to think through that some more

Well, the "rule" for `densityof` is that it's relative to the measures "natural" root measure, which in these cases should be Lebesgue. But we could make it explicit and use `density_rel`.

cscherrer:
Thinking about this some more... An SMF $F$ of a measure $\mu$ is a right-continuous function $\mathbb{R} \to \mathbb{R}$ so that 

$$
\mu((a, b]) = F(b) - F(a) = \lambda((F(a), F(b)])
$$

where $\lambda$ is Lebesgue measure. I guess the big question is, what's special about Lebesgue measure here? It's very nice computationally, because finding the different $F(b) - F(a)$ is so efficient.

OTOH, the implicit root measure in the `densityof` in
```julia
function smf(μ, x::Dual{TAG}) where TAG
    val = ForwardDiff.value(x)
    Δ = ForwardDiff.partials(x)
    Dual{TAG}(smf(μ, val), Δ * densityof(μ, val))
end
```
does *not* need to be Lebesgue. For example, everything works just fine for `Dirac(0.0)`, but the root measure is `CountingBase()`.

@mschauer do you have any insights on this?

cscherrer:
@oschulz maybe for now we stick with

$$
\mu((a, b]) = F(b) - F(a)
$$

but keep a lookout for benefits to generalizing this? I don't want to end up stalling our progress

oschulz:
> Thinking about this some more... An SMF F of a measure μ is a right-continuous function R→R so that
> 
> μ((a,b])=F(b)−F(a)=λ((F(a),F(b)])
> 
> where λ is Lebesgue measure. I guess the big question is, what's special about Lebesgue measure here?

Hm, maybe there's nothing special about the Lebesgue measure here, it's just that you can write any difference between numbers as Lebesgue measures. My intuition would be that the μ((a,b])=F(b)−F(a) is the central concept, and that it is based on the structure imposed by the sigma algebra. Is that correct?

src/smf.jl
```diff
@@ -0,0 +1,37 @@
...
+usually be called as `smf(μ, static(-Inf))`. It's then easy to ensure speed and
+avoid complex control flow by adding a method `smf(μ::M, ::StaticFloat64{-Inf})`.
+
+Users who pronounce `sinh` as "sinch" are advised to pronounce `smf` as "smurf".
+"""
+function smf end
+
+export smf
+
+function smfinv end
```
oschulz, line 27: Maybe `invsmf` instead?

src/standard/stdnormal.jl
```diff
@@ -12,7 +12,13 @@ export StdNormal
 
 @inline getdof(::StdNormal) = static(1)
 
-@inline transport_def(::StdUniform, μ::StdNormal, x) = erfc(-x * invsqrt2) / 2
-@inline transport_def(::StdNormal, μ::StdUniform, p) = -erfcinv(2 * p) * sqrt2
-
 @inline Base.rand(rng::Random.AbstractRNG, ::Type{T}, ::StdNormal) where {T} = randn(rng, T)
+
+Φ(z) = erfc(-z * invsqrt2) / 2
```
oschulz, line 17: Why the additional `Φ` function?

src/transport.jl
```diff
@@ -113,7 +120,7 @@ See [`transport_to`](@ref).
 """
 function transport_def end
 
-function transport_def(ν, μ, x)
+function transport_def(ν::NU, μ::MU, x::X) where {MU,NU,X}
```
oschulz, line 123: Do we need to add type parameters here?

cscherrer: No, that's vestigial

cscherrer, line 123:
```suggestion
function transport_def(ν, μ, x)
```

oschulz: I'd say let's take it out then.

oschulz: Ah, already gone :-)

src/smf.jl
cscherrer: I'm open to being convinced, but I like `smfinv` better. I haven't thought of *why* I like it better. Maybe
1. It's nice for things to start with what's more informative, for mental processing and for tab completion
2. Such close "S" sounds makes "inverse SMF" awkward to pronounce
3. An inverse function is $f^{-1}$ which we read "f inverse"; putting the "inverse" qualifier afterward is pretty common

src/standard/stdnormal.jl
cscherrer: Just a matter of breaking things down into irreducible components. We should probably also add methods for `InverseFunctions.inverse`. Also, if we have
```julia
smf(mu) = Base.Fix1(smf, mu)
```
then we can make this one a little nice:
```julia
smf(::StdNormal) = Φ
```
I think it should also make it a little simpler to add autodiff methods

oschulz: Sounds good.

src/smf.jl
cscherrer: Ok, we can change it

oschulz (approved):

PR #107, opened 2022-10-19
Dev

cscherrer:


src/interface.jl
```diff
@@ -27,7 +27,10 @@ using Test
 
 function dynamic_basemeasure_depth(μ::M) where {M}
     if static_hasmethod(proxy, Tuple{M})
-        return dynamic_basemeasure_depth(proxy(μ))
+        π = proxy(μ)
+        if static_hasmethod(basemeasure, Tuple{typeof(π)})
+            basemeasure(π) == basemeasure(μ) && return dynamic_basemeasure_depth(π)
+        end
```
cscherrer, line 33: Old code incorrectly assumed that any measure with a `proxy` method would be using that proxy to get the base measure. This update just checks that explicitly

oschulz (approved):

PR #108, opened 2022-10-24
Dev

cscherrer:


src/combinators/product.jl
```diff
@@ -212,10 +212,11 @@ end
 
 @inline function insupport(d::AbstractProductMeasure, x::AbstractArray)
     mar = marginals(d)
-    for (j, mj) in enumerate(mar)
-        dynamic(insupport(mj, x[j])) || return false
+    # We might get lucky and know statically that everything is inbounds
+    T = Core.Compiler.return_type(insupport, Tuple{eltype(mar),eltype(x)})
+    T == True || all(zip(x, mar)) do (xj, mj)
+        @inbounds insupport(mj, xj) == true
```
cscherrer, line 218:
```suggestion
        insupport(mj, xj) == true
```
No indexing, so we can drop the `@inbounds` here

src/combinators/product.jl
```diff
@@ -212,10 +212,11 @@ end
 
 @inline function insupport(d::AbstractProductMeasure, x::AbstractArray)
     mar = marginals(d)
-    for (j, mj) in enumerate(mar)
-        dynamic(insupport(mj, x[j])) || return false
+    # We might get lucky and know statically that everything is inbounds
+    T = Core.Compiler.return_type(insupport, Tuple{eltype(mar),eltype(x)})
+    T == True || all(zip(x, mar)) do (xj, mj)
```
cscherrer, line 217: No, it should be

```suggestion
    T <: True || all(zip(x, mar)) do (xj, mj)
```

oschulz (approved):

PR #115, opened 2023-05-30
Update to Static.jl v0.8

oschulz:
Closes #99 and #111 .

Since StaticInt is no longer an `Integer` in Static v0.8, we need to jump through a few hoops.

Drops support for older versions of Static.jl.

### DIFF
--- a/src/combinators/power.jl
+++ b/src/combinators/power.jl
@@ -17,8 +17,8 @@ struct PowerMeasure{M,A} <: AbstractProductMeasure
     axes::A
 end
 
-dslength(μ::PowerMeasure) = prod(dssize(μ))
-dssize(μ::PowerMeasure) = map(dslength, μ.axes)
+maybestatic_length(μ::PowerMeasure) = prod(maybestatic_size(μ))
+maybestatic_size(μ::PowerMeasure) = map(maybestatic_length, μ.axes)
 
 function Pretty.tile(μ::PowerMeasure)
     sz = length.(μ.axes)
--- a/src/static.jl
+++ b/src/static.jl
@@ -43,19 +43,19 @@ end
 end
 
 """
-    MeasureBase.dslength(x)::IntegerLike
+    MeasureBase.maybestatic_length(x)::IntegerLike
 
 Returns the length of `x` as a dynamic or static integer.
 """
-dslength(x) = length(x)
-dslength(x::AbstractUnitRange) = length(x)
-function dslength(::Static.OptionallyStaticUnitRange{StaticInt{A},StaticInt{B}}) where {A,B}
+maybestatic_length(x) = length(x)
+maybestatic_length(x::AbstractUnitRange) = length(x)
+function maybestatic_length(::Static.OptionallyStaticUnitRange{StaticInt{A},StaticInt{B}}) where {A,B}
     StaticInt{B - A + 1}()
 end
 
 """
-    MeasureBase.dssize(x)::Tuple{Vararg{IntegerLike}}
+    MeasureBase.maybestatic_size(x)::Tuple{Vararg{IntegerLike}}
 
 Returns the size of `x` as a tuple of dynamic or static integers.
 """
-dssize(x) = size(x)
+maybestatic_size(x) = size(x)


.github/workflows/CI.yml
```diff
@@ -6,8 +6,12 @@ on:
   push:
     branches:
       - master
+      - main
+      - dev
+      - 'releases/**'
```
theogf, line 11: I don't think you need those.

Project.toml
```diff
@@ -48,6 +48,6 @@ NaNMath = "0.3, 1"
 PrettyPrinting = "0.3, 0.4"
 Reexport = "1"
 SpecialFunctions = "2"
-Static = "0.5, 0.6"
+Static = "0.5, 0.6, 0.8"
```
theogf, line 51: Given your changes, shouild we just drop 0.5, 0.6 (and also what about 0.7)

src/combinators/power.jl
```diff
@@ -76,7 +85,7 @@ end
 end
 
 @inline function logdensity_def(
-    d::PowerMeasure{M,Tuple{Base.OneTo{StaticInt{N}}}},
+    d::PowerMeasure{M,Tuple{StaticOneTo{N}}},
```
theogf, line 88: StaticOneTo might be undefined no?

src/static.jl
```diff
@@ -0,0 +1,80 @@
...
+    # ranges, some operations that automatic differentiation requires do fail
+    # on such instances of `Fill` (e.g. `reshape` from dynamic to static size).
+    # So need to use standard ranges for the axes for now:
+    dyn_axs = map(_dynamic, axs)
+    FillArrays.Fill(x, dyn_axs)
+end
+
+
+"""
+    MeasureBase.dslength(x)::IntegerLike
```
theogf, line 67: The `ds_` naming is a bit hard to follow. Maybe there a could be another convention?

Project.toml
oschulz: 0.7 we can't support, static numbers aren't even `Real`s there. :-(

src/static.jl
oschulz: I thought about `maybestatic_length` but that's a bit of a mouthful. :-)

Any suggestions?

Project.toml
oschulz: I definitely want to keep 0.5 and 0.6 support for a little bit longer, though.

.github/workflows/CI.yml
oschulz: I can take that out though, if you prefer?

src/combinators/power.jl
oschulz: StaticOneTo is gone now.

src/combinators/power.jl
```diff
@@ -1,58 +1,67 @@
...
+"""
+    struct PowerMeasure{M,...} <: AbstractProductMeasure
 
-export PowerMeasure
+A power measure is a product of a measure with itself. The number of elements in
+the product determines the dimensionality of the resulting support.
 
+Note that power measures are only well-defined for integer powers.
+
+The nth power of a measure μ can be written μ^x.
```
cscherrer, line 13:
```suggestion
The nth power of a measure μ can be written μ^n.
```

.github/workflows/CI.yml
oschulz: I removed it, maybe it was a bit overkill :-)

src/combinators/power.jl
```diff
@@ -1,58 +1,67 @@
...
+"""
 struct PowerMeasure{M,A} <: AbstractProductMeasure
     parent::M
     axes::A
 end
 
+dslength(μ::PowerMeasure) = prod(dssize(μ))
+dssize(μ::PowerMeasure) = map(dslength, μ.axes)
```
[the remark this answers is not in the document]
oschulz: It stands for "dynamic or static". As @theogf remarked, probably not an ideal naming convention. Any ideas?

src/static.jl
cscherrer: This is just internal, right? If end users won't need to type it in the REPL, I think I'd lean toward making it more descriptive, even if it's longer. Most important for people reading the code to understand it

oschulz: Yes, it'll be mostly internal, though packages that use deeper measure features may need them. I've changed to the long names now.

cscherrer:
Just to be sure I understand, it seems like static floats are working again with our interface, but ints are not, so you added special cases and dispatch patterns to handle this. Is that right?

oschulz:
Ok now @cscherrer ?

cscherrer:
We should still check MeasureTheory tests locally. Latest MT has
```toml
Static = "0.5, 0.6"
```
Does that mean this PR is technically a breaking change?

cscherrer (approved):

oschulz:
> Ok in that case let's go with this and then MT can catch up

I'll see if I can do the Distributions extension soon, then MT can be updated in one go, maybe.

PR #116, opened 2023-05-30
Use StaticInteger instead of StaticInt and add tests for static util functions

oschulz:
Follow-up to #115 .

We should use `StaticInteger` instead of `StaticInt` for argument types.

Also adds tests for the new functions in static.jl.


oschulz:
Tests fixed.

theogf (approved):
Nice, it just needs a version bump

cscherrer (approved):

oschulz:
Added version bump.

oschulz:
Merci @theogf !

oschulz:
Next on the agenda will the polishing up the transport framework and turning DistributionMeasures.jl into a MeasureBase.jl Pkg-extension (DistributionMeasures.jl itself will become deprecated).

PR #119, opened 2023-06-22
Remove/rename some functionality and move operators into submodule

oschulz:
@cscherrer these is the first PR to implement the somewhat breaking changes we discussed over the last days (PRs to be collected on the mt-015 branch for a MeasureBase v0.15 release sometime next week).

Most of the removed/renamed functionality is rarely used, and the operators can be accessed via `using MeasureBase.MeasureOperators` now, so user code should be easy to adapt.

PR has a clean history, we should **not** squash it when we merge into mt-015, so we can keep track more easily.


src/combinators/bind.jl
```diff
@@ -3,31 +3,30 @@ struct Bind{M,K} <: AbstractMeasure
...
-probabilistic programming language like Soss.jl, this could be expressed as
+- a measure μ
+- a kernel function k that takes values from the support of μ and returns a
+  measure
 
-Note that bind is usually written `>>=`, but this symbol is unavailable in Julia.
+The *monadic bind* operation `mbind(k, μ)` returns is a new measure.
+
+A monadic bind ofen written as `>>=` (e.g. in Haskell), but this symbol is
+unavailable in Julia.
```
cscherrer, line 19:
```suggestion
The *monadic bind* operation `mbind(k, μ)` returns is a new measure.
If `ν == mbind(k, μ)` and all measures involved are sampleable, then 
samples from `rand(ν)` follow the same distribution as those from `rand(k(rand(μ)))`.


A monadic bind ofen written as `>>=` (e.g. in Haskell), but this symbol is
unavailable in Julia.
```

src/combinators/bind.jl
```diff
@@ -3,31 +3,30 @@ struct Bind{M,K} <: AbstractMeasure
...
+    pushfwd(Base.Fix1(*, scale), StdNormal())
 end
 ```
-
-See also `bind` and `Bind`
 """
-↣(μ, k) = bind(μ, k)
-
-bind(μ, k) = Bind(μ, k)
+mbind(k, μ) = Bind(μ, k)
```
cscherrer, line 28: Should we also swap the arguments in `Bind`?

src/combinators/likelihood.jl
```diff
@@ -91,34 +91,35 @@ Similarly to the above, we have
...
 
-``P(θ|x) ∝ P(θ) P(x|θ)``
+``P(θ|x) ∝ P(x|θ) P(θ)``
 
-The product on the right side is computed pointwise. To work with this in
-MeasureBase, we have a "pointwise product" `⊙`, which takes a measure and a
-likelihood, and returns a new measure, that is, the unnormalized posterior that
-has density ``P(θ) P(x|θ)`` with respect to the base measure of the prior.
+In measure theory, the product on the right side is actually the Lebesgue integral,
+of the likelihood with respect to the prior.
```
cscherrer, line 97:
```suggestion
In measure theory, the product on the right side is the Lebesgue integral
of the likelihood with respect to the prior.
```

src/combinators/likelihood.jl
```diff
@@ -91,34 +91,35 @@ Similarly to the above, we have
...
 struct Likelihood{K,X} <: AbstractLikelihood
     k::K
     x::X
 
-    Likelihood(k::K, x::X) where {K<:AbstractTransitionKernel,X} = new{K,X}(k, x)
-    Likelihood(k::K, x::X) where {K<:Function,X} = new{K,X}(k, x)
-    Likelihood(μ, x) = Likelihood(kernel(μ), x)
+    Likelihood(k::K, x::X) where {K,X} = new{K,X}(k, x)
+#!!!!!!!!!!!    # For type stability if `K isa UnionAll (e.g. a parameterized MeasureType)`
+    Likelihood(::Type{K}, x::X) where {K<:AbstractMeasure,X} = new{K,X}(K, x)
```
cscherrer, line 122: @cscherrer review this

src/measure_operators.jl
```diff
@@ -0,0 +1,141 @@
+"""
+    module MeasureOperators
+
+Defines the following operators for measures:
+
+* `f ⋄ μ == pushfwd(f, μ)`
+
+* `μ ∘ f == inverse(f) ⋄ μ`
```
cscherrer, line 8:
```suggestion
* `μ ⊙ f == inverse(f) ⋄ μ`
```

src/combinators/bind.jl
oschulz: Done.

cscherrer (dismissed):

PR #123, opened 2023-08-04
Fix `Vararg` warnings

cscherrer:
In Julia 1.10 beta (at least) we get this warning a lot:
```
│  WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
│  You may need to write `f(x::Vararg{T})` rather than `f(x::Vararg{<:T})` or `f(x::Vararg{T}) where T` instead of `f(x::Vararg{T} where T)`.
```

[Here](https://discourse.julialang.org/t/deprecation-warnings-from-atom-with-v1-7/72400/4?u=cscherrer) Simeon Schaub suggests that to address this, "Just use Vararg{Number} instead of Vararg{<:Number}". This PR makes this small change

### DIFF
--- a/src/combinators/power.jl
+++ b/src/combinators/power.jl
@@ -29,7 +29,7 @@ end
 
 # ToDo: Make rand return static arrays for statically-sized power measures.
 
-function _cartidxs(axs::Tuple{Vararg{<:AbstractUnitRange,N}}) where {N}
+function _cartidxs(axs::Tuple{Vararg{AbstractUnitRange,N}}) where {N}
     CartesianIndices(map(_dynamic, axs))
 end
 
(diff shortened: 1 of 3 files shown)


oschulz:
I have this in #122 already, I think - if it's ok I'd hold off until that done (hopefully very soon) to reduce merge conflicts.

cscherrer:
I don't think so, the pr targets that branch

cscherrer:
OTOH no problem waiting if you prefer, it's pretty minor I think

oschulz:
Thanks - I'll get back to that monster-PR, promise!

oschulz:
Included in #142

PR #139, opened 2024-01-27
Fix compats for MeasureBase and drop 1.3

theogf:
Due to the separation of the `stdlibs` with 1.9 we need compats on stdlibs as well, notably for the `Aqua` tests to pass


theogf:
I see the tests are only testing from 1.6, does that mean we should change the julia compat to 1.6 instead of 1.3?

oschulz:
Thanks @theogf !

How about we require Julia v1.6? Then we wouldn't need the `<0.0.1` compat-entries.

(NB: I'll see that I really get #122 ready soon).

oschulz (approved):

oschulz:
Ok, here we go: https://github.com/JuliaRegistries/General/pull/99684

PR #151, opened 2024-09-27
proxyfuns

cscherrer:
MeasureTheory.jl currently has functions like this:
```julia
StatsBase.entropy(m::AbstractMeasure, b::Real) = entropy(proxy(m), b)
Statistics.mean(m::AbstractMeasure) = mean(proxy(m))
Statistics.std(m::AbstractMeasure) = std(proxy(m))
Statistics.var(m::AbstractMeasure) = var(proxy(m))
Statistics.quantile(m::AbstractMeasure, q) = quantile(proxy(m), q)
```

But `proxy` is defined in MeasureBase, so this is type piracy. Better to move them here if possible.


oschulz:
I'd like to pull this into the other major changes I'm working on - the resulting version of MeasureBase will require Julia v1.10 (designated as the next LTS), so we can so this via an extension then.

PR #152, opened 2024-10-16
Random updates

cscherrer:
Just a placeholder for now, some of these changes could be rolled back or moved to MeasureTheory


oschulz:
Can we integrate this into #122 ?

oschulz:
> what's here is a side-effect of me working on some commercial code

Oh, ok - then let's get it in directly and I'll rebase #122.

PR #153, opened 2024-11-03
Add implicitly mapped measures and kernels

oschulz:
Also:

* Rename TransformVolCorr and subtypes (with backward compatibility).
* Improved pushforward measure implementation.
* Improve type stability of `logdensityof` for primitve, standard and combinator measures.


src/combinators/half.jl
```diff
@@ -19,6 +19,11 @@ function Base.rand(rng::AbstractRNG, ::Type{T}, μ::Half) where {T}
     return abs(rand(rng, T, unhalf(μ)))
 end
 
+function logdensityof(μ::Half, x)
+    ld = logdensityof(unhalf(μ), x) - loghalf
```
cscherrer, line 23: `loghalf` is undefined?

src/primitive.jl
```diff
@@ -19,6 +19,9 @@ basemeasure(μ::PrimitiveMeasure) = μ
 
 @inline basemeasure_depth(::PrimitiveMeasure) = static(0)
 
+@inline logdensityof(::PrimitiveMeasure, x::Real) = zero(float(typeof(x)))
+@inline logdensityof(::PrimitiveMeasure, x) = false
```
cscherrer, line 23: Do we need to convert this to a float? A little worried about type stability here

src/combinators/transformedmeasure.jl
```diff
@@ -29,13 +69,23 @@ export PushforwardMeasure
...
-struct PushforwardMeasure{F,I,M,VC<:TransformVolCorr} <: AbstractPushforward
+struct PushforwardMeasure{F,I,M,VC<:PushFwdStyle} <: AbstractPushforward
     f::F
     finv::I
     origin::M
     volcorr::VC
 end
 
+function PushforwardMeasure(f, finv, origin, volcorr)
+    PushforwardMeasure{Core.Typeof(f),Core.Typeof(finv)}(origin, volcorr)
```
cscherrer, line 80: Missing `f` and `finv` arguments here

src/combinators/half.jl
oschulz: `loghalf` is in `IrrationalConstants`.

src/combinators/spikemixture.jl
```diff
@@ -21,11 +21,17 @@ end
...
 
-@inline function logdensity_def(μ::SpikeMixture, x)
-    if iszero(x)
-        return log(μ.s)
-    else
-        return log(μ.w) + logdensity_def(μ.m, x)
+for func in [:logdensityof, :logdensity_def]
+    @eval @inline function $func(μ::SpikeMixture, x)
+        R1 = Core.Compiler.return_type(log, Tuple{typeof(μ.s)})
+        R2 = Core.Compiler.return_type(log, Tuple{typeof(μ.w)})
```
cscherrer, line 27: This seems a little complicated. If we changed it to
```julia
R1 = typeof(log(one(μ.s)))
R2 = typeof(log(one(μ.w)))
```
we could really less on type inference, and maybe constant propagation would still compile it away?

src/primitive.jl
oschulz: We might otherwise end up with an integer `0`. Which would be fine here, I guess, but in many other places we to need to do this so we can return either `zero(float(typeof(x)))` or `float(typeof(x))(-Inf)`. So I did it here as well for consistency. But it's absolutely type-stable.

src/combinators/transformedmeasure.jl
oschulz: But `f` and `finv` are function arguments?

src/combinators/half.jl
cscherrer: ahh ok, let's make the import explicit in the file so it's clearer

src/combinators/spikemixture.jl
oschulz: I don't trust constant propagation, esp. in combination with automatic differentiation. :-)

src/primitive.jl
cscherrer: Would static(0.0) have the same performance? If so that would be clearer to read

src/combinators/half.jl
oschulz: Done.

src/primitive.jl
oschulz: I tried `static(0.0)`. Unfortunately `(static(0.0) + Float32(1)) isa Float64` which throws a big wrench into future GPU-support plans (except for very expensive GPUs).

oschulz: I'm afraid we'll need to do less `static(floating_point_constant)` in the future, instead of more.

src/combinators/spikemixture.jl
cscherrer: Ok I've seen some issues with type inference (it's stateful and very hard to predict) but let's try this and come back if we see issues. I added a note

src/combinators/transformedmeasure.jl
cscherrer: I don't understand. When you call 
```julia
MyStruct{types}(args...)
```
`args` need to match the struct definition. Unless maybe this is forwarding to another constructor?

oschulz: Yes, it forwards to the default constructor. But the `Core.Typeof` is important, in case `f` is a ctor.

oschulz: In any case it passes tests :-)

src/primitive.jl
oschulz: Can I mark this resolved?

src/combinators/transformedmeasure.jl
oschulz: Fixed now.

oschulz:
Rebased PR, moved reformats into topical commit for better history, no change in total diff.

oschulz:
Require Julia v1.10 and added ChainRulesCore extension, so new rrules in this PR can go into the right place directly.

cscherrer (approved):

PR #155, opened 2024-11-09
[WIP] Move from implicitly mapped measures and kernels to data tagged as mapped

oschulz:


src/combinators/half.jl
```diff
@@ -19,8 +19,8 @@ function Base.rand(rng::AbstractRNG, ::Type{T}, μ::Half) where {T}
     return abs(rand(rng, T, unhalf(μ)))
 end
 
-function logdensityof(μ::Half, x)
-    ld = logdensityof(unhalf(μ), x) - loghalf
+function strict_logdensityof(μ::Half, x)
+    ld = strict_logdensityof(unhalf(μ), x) - loghalf
```
cscherrer, line 23: I need to have another look at this - should it just be `logdensity_def`?

If we need to merge before this is resolved, let's add a #TODO comment

oschulz: No, this is part of specializing logdensityof (https://github.com/JuliaMath/MeasureBase.jl/pull/153/commits/e24231465905534bafcd5f2c02a063a085cb7846) to avoid the multi-step `logdensity_def` machinery if users don't need `logdensity_rel` (until we revamp that machinery to make it more type-stable and Zygote-friendly).

PR #160, opened 2025-07-25
Extend Static/StaticArrays tooling and adapt PowerMeasure

oschulz:
This contains all changes for static tools that lived on the mt-15 branch - now updated, improved and with a lot of tests. I'll provide the basis for generating and handling static array variates and so on.

Slightly breaking, but we have accumulated a few breaking changes on main already.

### DIFF
--- a/src/MeasureBase.jl
+++ b/src/MeasureBase.jl
@@ -31,6 +31,7 @@ import ConstructionBase
 using ConstructionBase: constructorof
 using IntervalSets
 
+import StaticArrays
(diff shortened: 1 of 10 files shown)


oschulz:
The static tools have grown quite extensive and general, I'll probably pull them out into a separate package, part of this PR  may go away again soon.

oschulz:
Static tools will live in https://github.com/oschulz/StaticThings.jl, being registered now, will upate this PR when registration is complete.

